PR #22, opened 2019-10-27
Add Resample and Downsample transforms

### LINKED ISSUE #6: Add downsampling transformation

askskro:
Add Resample and Downsample transforms with tests and demo usage.
Closes #6 
Feedback requested on default values for min and max sampling range.


iver56:
max_sample_rate=44100 (CD quality) seems like an apt default

iver56:
I think we can remove `Downsample` and make do with `Resample`. We could add to the `Resample` docstring that it can be used for downsampling by specifying small min_sample_rate and max_sample_rate

askskro:
Thanks for the feedback. Changes made.

iver56:
Thanks! ⭐️ 🥇

PR #23, opened 2019-10-30
Add Distortion transformation

### LINKED ISSUE #5: Add distortion transformation

askskro:
Adds a very simple distortion transformation.

Closes #5

### DIFF
--- a/audiomentations/augmentations/transforms.py
+++ b/audiomentations/augmentations/transforms.py
@@ -277,24 +277,33 @@ class Resample(BasicTransform):
 
 
 class ClippingDistortion(BasicTransform):
-    """Distort signal by clipping the signal at a threshold"""
+    """Distort signal by clipping a random percentage of points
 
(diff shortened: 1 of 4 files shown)


iver56:
Thanks for the contribution!

Here are my thoughts thus far: It's not obvious from the parameter name docstring that the actual percentile is selected randomly. Perhaps it can be made more obvious by having _two_ parameters min_percentile_threshold and max_percentile_threshold instead of a single parameter percentile_cut_off. I would rather call it threshold than cutoff to avoid confusion, since cutoff is usually a term used for a parameter you can tweak to change the frequency response of a low-pass/high-pass filter.

iver56:
Also, perhaps it's best to rename the class to `ClippingDistortion`, since many other kinds of distortion also exist?

askskro:
Thanks for the feedback. Suggested changes have been made.

iver56:
Nice!

Issue #37, opened 2020-04-02
Add a transformation that adds mp3 or vorbis compression artifacts


iver56:

iver56:
I'm putting this on hold until https://github.com/bastibe/SoundFile/issues/130 has been fixed

iver56:
Maybe I can use https://github.com/jiaaro/pydub instead. It has support for mp3 export, given that ffmpeg is installed.

PR #38, opened 2020-04-07
Implement Mp3Compression transform

iver56:
It's not really ideal, since it writes to disk, but at least it is functional.

Close https://github.com/iver56/audiomentations/issues/37

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,9 +14,9 @@ A Python library for audio data augmentation. Inspired by [albumentations](https
 
 ## Optional requirements
 
-The `Mp3Compression` transform depends on `pydub` and `ffmpeg`. The `pydub` dependency can be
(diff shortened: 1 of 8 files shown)


iver56:
This approach may be better:
https://github.com/facebookresearch/demucs/commit/7157c0879922b7d9d72f869bd4c4fd05abb945e3#diff-1c77eb608c381bd9ce78849390fd6311R82

iver56:
Actually, I think we should support both backends. Pros/cons:

# lameenc:

Pros: More flexible parameters. Faster than pydub/ffmpeg.

Cons: Introduces latency in the audio

# pydub/ffmpeg

Pros: Does not introduce latency in the audio

Cons: Slower than lameenc

PR #39, opened 2020-05-03
Fix bug in frequency mask

kvilouras:
sosfilt() function eliminates NaN values (they're now 0 instead). Also, i fixed the upper bound of the "freq_start" parameter. For example, let's take the worst case scenario ("bandwidth" and "freq_start" equal to their upper bound). That means that highcut will be equal to the Nyquist frequency.
SInce the filter's critical frequencies should be 0 < Wn <1, we are now making sure that this type of error will never occur.

### DIFF
--- a/audiomentations/augmentations/transforms.py
+++ b/audiomentations/augmentations/transforms.py
@@ -77,12 +77,12 @@ class FrequencyMask(BasicTransform):
         nyq = 0.5 * fs
         low = lowcut / nyq
         high = highcut / nyq
-        b, a = butter(order, [low, high], btype="bandstop")
-        return b, a
+        sos = butter(order, [low, high], btype='bandstop', output='sos')
+        return sos
 
     def __butter_bandstop_filter(self, data, lowcut, highcut, fs, order=5):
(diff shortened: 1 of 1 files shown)


iver56:
Thanks! This does indeed seem to fix the bug. I have added a test https://github.com/iver56/audiomentations/commit/defce915f796394aef3e51e72bff18da8b63baff that fails before this patch and succeeds after this patch is applied

A few minor changes are needed before I'll merge this:
* `sosfilt` needs to be imported (and remove `lfilter` from the list of things to import)
* Format the code with `black` (or manually replace `'` with `"`)

kvilouras:
> Thanks! This does indeed seem to fix the bug. I have added a test [defce91](https://github.com/iver56/audiomentations/commit/defce915f796394aef3e51e72bff18da8b63baff) that fails before this patch and succeeds after this patch is applied
> 
> A few minor changes are needed before I'll merge this:
> 
> * `sosfilt` needs to be imported
> * Remove `lfilter` from imports
> * Format the code with `black` (or, in this case, manually replace `'` with `"`)

sorry about the imports. glad i could help.

PR #42, opened 2020-05-06
save wav files as float32 format

xwJohn:
Some app can't open float64 format wav file, such as Window Media Player and Praat.


iver56:
Thanks :)

Quick thought: `np.divide(a, b, dtype=np.float32)` might be apt here

PR #67, opened 2020-11-06
Configurable LRU cache for transforms that use external sound files

### LINKED ISSUE #45: Configurable cache size
Today, the LRU cache for loading sounds is hardcoded to max 64 sounds in AddShortNoises. Let's make it configurable, because the caching needs vary from case to case.

alumae:
Fixes #45


iver56:
Thanks!

PR #72, opened 2020-11-24
Implement `Compose.randomize_parameters`. Close #71

### LINKED ISSUE #71: Add `randomize_parameters()` in composition
Hi!
I guess it would be nice to use freezing of parameters before all transformation right after generating them.
```python
augment = Compose([...])

...

# in any training loop

waves_to_equal_augment = [...]
sr = ...

augment.randomize_parameters(waves_to_equal_augment[0], sr)
augment.freeze_parameters() # it can be even used once!
for wav in waves_to_equal_augment:
  aug_wav = augment(wav)
  ...

augment.unfreeze_parameters() # it can be even dropped!
```

Currently parameters even can't be generated without `apply` of composition.

SolomidHero:
Decided in #71 

Also test added on this method in `tests/test_compose.py`

### DIFF
--- a/audiomentations/core/composition.py
+++ b/audiomentations/core/composition.py
@@ -22,6 +22,13 @@ class Compose:
 
         return samples
 
+    def randomize_parameters(self, samples, sample_rate):
+        """
+        Randomise and define parameters of every transform in composition.
+        """
+        for transform in self.transforms:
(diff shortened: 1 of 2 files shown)


iver56:
Thanks!

Issue #66, opened 2020-11-05
Implement SpecCompose


iver56:

omerferhatt:
#80 PR added

omerferhatt:
Added to master, this issue closed now

iver56:
There seems to be a few failures in the test suite.

```
============================================================================================== FAILURES ==============================================================================================
________________________________________________________________________ TestSpecCompose.test_freeze_and_unfreeze_parameters _________________________________________________________________________

self = <tests.test_spec_compose.TestSpecCompose testMethod=test_freeze_and_unfreeze_parameters>

    def test_freeze_and_unfreeze_parameters(self):
        spectrogram = np.zeros((256, 256), dtype=np.float32)
        augmenter = SpecCompose(
            [
                SpecChannelShuffle(),
                SpecFrequencyMask(),
            ]
        )
>       perturbed_samples1 = augmenter(magnitude_spectrogram=spectrogram)

tests\test_spec_compose.py:22:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
audiomentations\core\composition.py:71: in __call__
    magnitude_spectrogram = transform(magnitude_spectrogram)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <audiomentations.augmentations.spectrogram_transforms.SpecChannelShuffle object at 0x000001B64155BEB8>
magnitude_spectrogram = array([[0., 0., 0., ..., 0., 0., 0.],
       [0., 0., 0., ..., 0., 0., 0.],
       [0., 0., 0., ..., 0., 0., 0.],
[...]

PR #80, opened 2021-01-08
SpecCompose Implementation

omerferhatt:
- BaseCompose class added
- Compose class inherited from BaseCompose
- SpecCompose class created according to magnitude spectrogram
- SpecCompose unit test added

### DIFF
--- a/audiomentations/__init__.py
+++ b/audiomentations/__init__.py
@@ -19,6 +19,6 @@ from .augmentations.transforms import (
     TimeStretch,
     Trim,
 )
-from .core.composition import Compose
+from .core.composition import Compose, SpecCompose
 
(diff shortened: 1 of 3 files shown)


omerferhatt:
According to #66 issue

iver56:
Thanks for the PR! I'll have a look

iver56:
The only remark I have is that we need to import `SpecCompose` in audiomentations/\_\_init\_\_.py. `Compose` is also imported there. This makes the class easily importable for people who have install the audiomentations package.

omerferhatt:
I added to __init__ in audiomentations. is there anything that I need to fix ?

iver56:
Thanks

PR #82, opened 2021-01-11
BugFix on SpecChannelShuffle

omerferhatt:
- SpecChannelShuffle channel fixed channel count changed to dynamic.
- Test sound changed to random gaussian noise with multichannel in test_spec_compose.py
- Multichannel spectrogram check function updated


iver56:
Thanks!

PR #90, opened 2021-02-23
add support for 32-bit loading in the load_wav_file function in

fmirus:
add 32-bit support for loading wav files in audio_loading_utils.py


iver56:
Thanks!

iver56:
By the way, I would like to back this up with two unit tests.

The thing is that both 32-bit and 24-bit wavs end up having dtype int32, and they need to be divided by different integers

iver56:
Would you like to write a test or two?

iver56:
https://github.com/iver56/audiomentations/blob/master/tests/test_load_sound_files.py

PR #91, opened 2021-03-03
Add a fade option in the Shift transform

mmxgn:
* Add a `fade` and a `fade_duration` parameter in
`audiomentations.augmentations.transforms.Shift`
* Add the corresponding tests

There are cases where a shift transform without rollover would introduce abrupt changes from zero to non-zero samples leading to transients such as ones in the following image:

image

For this reason I have made the following pull request which adds a `fade: bool` and a `fade_duration` `int` or `float` to `Shift.__init__`. When `fade_duration` is an int the fade is in samples while when its a `float` it is a time duration.

### DIFF
--- a/audiomentations/augmentations/transforms.py
+++ b/audiomentations/augmentations/transforms.py
@@ -350,7 +350,7 @@ class Shift(BaseWaveformTransform):
         max_fraction=0.5,
         rollover=True,
         fade=False,
-        fade_duration=10,
+        fade_duration=0.01,
         p=0.5,
     ):
         """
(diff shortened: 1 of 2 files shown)


audiomentations/augmentations/transforms.py
```diff
@@ -374,11 +390,31 @@ def randomize_parameters(self, samples, sample_rate):
...
+                    shifted_samples[
+                        ...,
+                        num_places_to_shift - 1 : num_places_to_shift + fade_length - 1,
+                    ] *= np.linspace(0, 1, num=fade_length)
             elif num_places_to_shift < 0:
                 shifted_samples[..., num_places_to_shift:] = 0.0
+                if self.fade:
+                    shifted_samples[
+                        ...,
+                        num_places_to_shift - fade_length + 1 : num_places_to_shift + 1,
```
iver56, line 416: Can you make a unit test that checks if this breaks if fade_length is very large compared to the length of the audio? I'm not sure what happens when a negative index value is less than `-len(the_array)`

audiomentations/augmentations/transforms.py
```diff
@@ -374,11 +390,31 @@ def randomize_parameters(self, samples, sample_rate):
...
             if num_places_to_shift > 0:
                 shifted_samples[..., :num_places_to_shift] = 0.0
+                if self.fade:
+
+                    # When fading in, the first sample is always 0, that's why
+                    # the fade_length - 1 below.
+
+                    shifted_samples[
+                        ...,
+                        num_places_to_shift - 1 : num_places_to_shift + fade_length - 1,
```
iver56, line 409: Can you check (with a unit test) if this breaks when fade_length is very large?

iver56 (changes requested):

iver56:
Shift _with_ rollover can also lead to a transient. Should we fix that in a future pull request?

audiomentations/augmentations/transforms.py
mmxgn: I did in my new commit.

audiomentations/augmentations/transforms.py
```diff
@@ -344,22 +344,38 @@ class Shift(BaseWaveformTransform):
...
         """
         :param min_fraction: float, fraction of total sound length
         :param max_fraction: float, fraction of total sound length
         :param rollover: When set to True, samples that roll beyond the first or last position
             are re-introduced at the last or first. When set to False, samples that roll beyond
             the first or last position are discarded. In other words, rollover=False results in
             an empty space (with zeroes).
+        :param fade: When set to True, and if rollover is False, there will be a short fade
+            in and out of the silent of duration `fade_duration`.
+        :param fade_samples: The duration of the fade. If it is an integer, then it refers to
```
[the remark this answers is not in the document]
mmxgn: Changed to accept always seconds in duration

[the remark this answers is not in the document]
iver56: Nice :) Could you also please update the docstring so it says fade_duration instead of fade_samples?

iver56 (approved):

iver56:
Thanks! :)

PR #92, opened 2021-03-05
Fixed code quality issues

withshubh:
#### Description
Hi :wave: 
I ran the [DeepSource](https://deepsource.io) analysis on the forked copy of this repo and found some interesting [code quality issues](https://deepsource.io/gh/withshubh/audiomentations/issues/?category=recommended).

#### Summary of changes

- Remove methods with unnecessary super delegation.
- Refactor unnecessary `else` / `elif` when `if` block has a `return` statement
- Change methods not using its bound instance to staticmethods
- Remove implicit `object` from the base class
- Add .deepsource.toml

### DIFF
--- a/.deepsource.toml
+++ b/.deepsource.toml
@@ -1,0 +1,10 @@
+version = 1
+
+test_patterns = ["tests/test_*.py"]
+
+[[analyzers]]
+name = "python"
+enabled = true
+
+  [analyzers.meta]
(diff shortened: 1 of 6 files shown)


iver56:
I feel like this is an automated ad for the company you work for and that these "code quality issues" are either insignificant or make the code slightly worse. Like removing a TODO without addressing it and marking a method as static while it shouldn't be static.

PR #93, opened 2021-03-08
Feature/24 bit support

fmirus:
Now also add 24-bit support making use of both scipy.io and wavio read functions. This is able to read almost all wav files. I still get some errors with 24-bit files with very high sampling rates (order of magnitude 44k), which seems to be a limitation in the wavio/wave read functions themselves.

### DIFF
--- a/audiomentations/core/audio_loading_utils.py
+++ b/audiomentations/core/audio_loading_utils.py
@@ -4,6 +4,7 @@ from pathlib import Path
 import librosa
 import numpy as np
 from scipy.io import wavfile
+from wavio import read
 
 IS_WAVIO_INSTALLED = True
 try:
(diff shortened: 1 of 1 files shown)


iver56:
* I would really like to have some tests for this. It would be nice to have some small test fixture files with various formats.
* There is a function called `load_wav_file_with_wavio`. What should we do with it? Having duplicated code feels suboptimal
* I think the newest scipy loads 24-bit wavs with `samples.dtype == np.int32`, so it would go into the same part of the code as the int32 files, where we divide by 2147483648. How should we handle that?

audiomentations/core/audio_loading_utils.py
```diff
@@ -79,19 +80,44 @@ def load_wav_file(file_path, sample_rate, mono=True, resample_type="kaiser_best"
...
+        if samples.dtype != np.float32:
+            if samples.dtype == np.int16:
+                samples = np.true_divide(
+                    samples, 32768, dtype=np.float32
+                )  # ends up roughly between -1 and 1
+            elif samples.dtype == np.int32:
+                samples = np.true_divide(
+                    samples, 2147483648, dtype=np.float32
+                )  # ends up roughly between -1 and 1
+    except:
```
iver56, line 94: Too broad exception handler. What kind of exception are we expecting here?

audiomentations/core/audio_loading_utils.py
```diff
@@ -79,19 +80,44 @@ def load_wav_file(file_path, sample_rate, mono=True, resample_type="kaiser_best"
...
+                # uint24
+                samples = np.true_divide(
+                    samples, 8388608 , dtype=np.float32
+                )  # ends up roughly between -1 and 1
+            elif sampwidth == 4:
+                # uint32
+                samples = np.true_divide(
+                    samples, 2147483648, dtype=np.float32
+                )  # ends up roughly between -1 and 1
+        except:
```
iver56, line 120: Too broad exception handler. For example, if the file no longer exists or cannot be read for some reason, we don't want to say "unexpected data type"

iver56 (changes requested):

iver56:
> I still get some errors with 24-bit files with very high sampling rates

Those would be interesting to have as test cases, if it is feasible to add them as small test fixture files

fmirus:
@iver56 Those are really good remarks. I was not aware, that the newest scipy supports all wav formats and thus basically solves this issue. Therefore, I withdraw this PR.

iver56:
Ok. I'll have a look myself to see how these issues can be fixed in a good way.
Thanks for the effort though.

PR #98, opened 2021-04-22
Fix a bug in AddImpulseResponse when samples = zero signal.

juice500ml:
Zero signal will yield max_value to be 0, hence divide by zero error will occur. This commit avoids the bug.


iver56:
Thanks

Issue #61, opened 2020-10-08
Implement SomeOf and OneOf


iver56:
Inspired by imgaug:

https://github.com/aleju/imgaug

https://github.com/aleju/imgaug/blob/0101108d4fed06bc5056c4a03e2bcb0216dac326/imgaug/augmenters/meta.py#L3188

omerferhatt:
I'll try to implement

iver56:
Thanks!

iver56:
https://github.com/iver56/audiomentations/pull/139

PR #99, opened 2021-05-07
add option to apply only one transformation to compose

Edresson:
To avoid extreme audio degradation due to overlapping transforms, we often want only one of the transformations to be applied within the Compose sequence.

### DIFF
--- a/audiomentations/core/composition.py
+++ b/audiomentations/core/composition.py
@@ -2,10 +2,11 @@ import random
 
 
 class BaseCompose:
-    def __init__(self, transforms, p=1.0, shuffle=False):
+    def __init__(self, transforms, p=1.0, shuffle=False, only_one=False):
         self.transforms = transforms
         self.p = p
         self.shuffle = shuffle
+        self.only_one = only_one
 
         name_list = []
         for transform in self.transforms:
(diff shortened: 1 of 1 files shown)


iver56:
Thanks for the contribution! It relates to https://github.com/iver56/audiomentations/issues/61

Ideally, I would like `OneOf` to be a separate class instead of being a parameter in `Compose`. This is inspired by imgaug. The code there can look somewhat like this:

```
seq = Sequential(
    [
        sometimes(
            Affine(
                # ...
            )
        ),
        # execute 0 to 5 of the following (less important) augmenters per image
        # don't execute all of them, as that would often be way too strong
        SomeOf(
            (0, 5),
            [
                # Apply exactly one kind of blur
                OneOf(
                    [
                        GaussianBlur((0, 3.0)),
                        AverageBlur(k=(2, 7)),
                        MedianBlur(k=(3, 11)),
                    ]
                ),
                Sharpen(alpha=(0, 1.0), lightness=(0.75, 1.5)),  # sharpen images
                OneOf(
                    [
                        Dropout((0.01, 0.1), per_channel=0.5),
                        CoarseDropout(
                            (0.03, 0.15), size_percent=(0.02, 0.05), per_channel=0.2
                        ),
                    ]
                ),
            ],
        ),
    ],
    random_order=True,
)
```

Can you make it happen? `SomeOf` is also desirable to have at some point

iver56:
Hi @Edresson 

Would you like to address the feedback? Or should I close this pull request?

iver56:
Closing due to inactivity

iver56:
OneOf is now available in master, and will become available in a future release

PR #102, opened 2021-05-30
support np.float64 in load_wav_file()

### LINKED ISSUE #101: AddBackgroundNoise doesn't work with np.float64 wav files
`AddBackgroundNoise `doesn't work with `np.float64` wav files. It shows `Unexpected data type` error. I've managed to fix this by converting samples to `np.float32` in `load_wav_file` function. Can you fix that?

jeongyoonlee:
This PR fixes #101 

- convert np.float64 to np.float32 in load_wav_file()


iver56:
Thanks

Issue #50, opened 2020-07-27
Add tanh distortion transform


iver56:

pranav6670:
Can I work on this?

iver56:
Yes please 👍

pranav6670:
Hey @iver56, can you educate me on where to add my code and writing tests! Sorry, I am kinda noob here:(

iver56:
You need to add a class `TanhDistortion` in https://github.com/iver56/audiomentations/blob/master/audiomentations/augmentations/transforms.py

You can look at other classes for inspiration

You also need to add a test file in this folder: https://github.com/iver56/audiomentations/tree/master/tests

You can look at the existing test files for inspiration

If you want to listen to the doings of your new augmentation transform, you can add it to the demo script and run it on your computer:

https://github.com/iver56/audiomentations/blob/master/demo/demo.py

pranav6670:
Sure, thank you for the directions. I am planning to add tanh, atan and exponential distortions.

iver56:
@pranav6670 How is it going?

atamazian:
Can you specify desired input arguments names?

atamazian:
Added with https://github.com/iver56/audiomentations/pull/107, please close this issue.

PR #107, opened 2021-06-11
Add TanhDistortion

atamazian:
Adds TanhDistortion according to https://github.com/iver56/audiomentations/issues/50

### DIFF
--- a/audiomentations/augmentations/transforms.py
+++ b/audiomentations/augmentations/transforms.py
@@ -1081,8 +1081,7 @@ class TanhDistortion(BaseWaveformTransform):
     def apply(self, samples, sample_rate):
         if self.parameters["max_amplitude"] > 0:
             distorted_samples = np.tanh(self.parameters["gain"]*samples)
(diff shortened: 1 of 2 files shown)


iver56:
Thanks for the initiative :) This is a good start. Here are some thoughts on how it can be improved further:

* Have a mechanism that randomly varies `c2` between a specified min value and max value
* Remove the `c1` parameter and have it calculated automatically on the fly so that the RMS of the input and the output becomes the same (AKA post gain / gain compensation)

What do you think?

atamazian:
Your suggestions look quite good, I'll try to implement them.

atamazian:
I've implemented all things you mentioned, however, I wasn't able to make RMS of transformed and original signal completely equal - they're equal if rounded up to 3 decimals.

iver56:
LGTM. Thanks for the contribution :) Note that I may do some minor changes later, some time before a new version of audiomentations with the new transform gets released

PR #108, opened 2021-06-13
Fix SNR inverse characteristics in AddGaussianSNR

atamazian:
Gaussian noise is now generated with `1/noise_std ` instead of `noise_std`.

### DIFF
--- a/audiomentations/augmentations/transforms.py
+++ b/audiomentations/augmentations/transforms.py
@@ -228,8 +228,8 @@ class AddGaussianSNR(BaseWaveformTransform):
             std = np.std(samples)
             if self.min_SNR is not None and self.max_SNR is not None:
                 if self.min_snr_in_db is not None and self.max_snr_in_db is not None:
-                    raise Exception("Use either min_SNR and max_SNR parameters, or use min_snr_in_db and" \ 
-                                    "max_snr_in_db parameters instead. Simultaneous usage of both of them" \ 
+                    raise Exception("Use either min_SNR and max_SNR parameters, or use min_snr_in_db and" \
+                                    "max_snr_in_db parameters instead. Simultaneous usage of both of them" \
                                     "is not allowed.")
                 else:
                     min_snr = self.min_SNR
--- a/audiomentations/augmentations/transforms.py
+++ b/audiomentations/augmentations/transforms.py
@@ -198,8 +198,7 @@ class AddGaussianSNR(BaseWaveformTransform):
                  max_snr_in_db=30,
                  min_SNR=None, 
                  max_SNR=None,
-                 p=0.5, 
-                 snr_characteristics=None
+                 p=0.5
      ):
         """
         :param min_snr_in_db: Minimum signal-to-noise ratio in db
@@ -210,13 +209,6 @@ class AddGaussianSNR(BaseWaveformTransform):
         :snr_characteristics: Use legacy mode or not
         """
         super().__init__(p)
-        self.snr_characteristics = snr_characteristics
-        if self.snr_characteristics is None:
-            warnings.warn(
-                'Please set snr_characteristics to either "legacy" or "correct".'
-                ' The default behavior may change from "legacy" to "default" in a later release'
-            )
-            self.snr_characteristics="legacy"
         self.min_snr_in_db = min_snr_in_db
         self.max_snr_in_db = max_snr_in_db
         self.min_SNR = min_SNR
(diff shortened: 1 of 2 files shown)


iver56:
Thanks for fixing this :)

I propose that we apply this breaking change in a "gradual" way: The code will continue working as it does for existing users, but they will get a warning if they don't actively set a parameter snr_characteristics in `__init__`. If they actively set the parameter to "correct", they will get the new, correct behavior. We could set the default value to "correct" later. Something like:

```
__init__(..., snr_characteristics=None)

self.snr_characteristics = snr_characteristics
if self.snr_characteristics is None:
    warnings.warn(
        'Please set snr_characteristics to either "legacy" or "correct".'
        ' The default behavior may change from "legacy" to "default" in a later release'
    )
    self.snr_characteristics="legacy"
```

Inside `apply`:

```
if self.snr_characteristics == "legacy":
    # use old code, which people may currently rely on
else:
    # use the new and more correct code
```

This way of gradually applying breaking changes is inspired by scikit-learn

iver56:
Furthermore, it would be nice to have the min/max snr parameters in decibels instead, but that could be done later

atamazian:
OK, I'll change the code accordingly.

atamazian:
The `AddGaussianSNR` function has now `min_snr_in_db` and `max_snr_in_db` arguments, which allow you to set SNR in db of added Gaussian noise correctly. If you have old parameters set, it will warn you to switch to `min_snr_in_db` and `max_snr_in_db` parameters.

atamazian:
And if you use legacy parameters, the bug remains. You can change that if you like. I've decided not to use `snr_characteristics`, and change function behavior depending on presence of new and old parameters.

audiomentations/augmentations/transforms.py
```diff
@@ -193,32 +193,69 @@ class AddGaussianSNR(BaseWaveformTransform):
 
     supports_multichannel = True
 
-    def __init__(self, min_SNR=0.001, max_SNR=1.0, p=0.5):
+    def __init__(self, 
+                 min_snr_in_db=3,
```
iver56, line 197: Some people may use this class like `AddGaussianSNR(0.01, 0.001)`. With this new parameter ordering, those numbers would be interpreted as db although they are not. Perhaps it's better to swap the order of parameters, so it doesn't become a breaking change (for now)?

i.e.
```
min_SNR=0.001,
max_SNR=1.0
min_snr_in_db=None,
max_snr_in_db=None,
```

(but people should still get the warning if they use the old parameters)

audiomentations/augmentations/transforms.py
```diff
@@ -193,32 +193,69 @@ class AddGaussianSNR(BaseWaveformTransform):
...
+        :param max_SNR: Maximum signal-to-noise ratio (legacy)
         :param p: The probability of applying this transform
+        :snr_characteristics: Use legacy mode or not
         """
         super().__init__(p)
+        self.snr_characteristics = snr_characteristics
+        if self.snr_characteristics is None:
+            warnings.warn(
+                'Please set snr_characteristics to either "legacy" or "correct".'
+                ' The default behavior may change from "legacy" to "default" in a later release'
```
iver56: "legacy" to _"correct"_

audiomentations/augmentations/transforms.py
atamazian: I've fixed that.

iver56: Another alternative would be to enforce keyword-only arguments (instead of allowing both positional and keyword arguments). But if we go down that path, we would maybe have to do the same in all other transforms as well for consistency, so I don't think that's the best solution

atamazian: I think that change of arguments order is the best choice for now.

audiomentations/augmentations/transforms.py
atamazian: I think that there's no need of separate `snr_characteristics` argument.

audiomentations/augmentations/transforms.py
```diff
@@ -193,34 +193,67 @@ class AddGaussianSNR(BaseWaveformTransform):
...
-        TODO: This should be fixed in a future release.
-
-        :param min_SNR: Minimum signal-to-noise ratio
-        :param max_SNR: Maximum signal-to-noise ratio
+        :param min_snr_in_db: Minimum signal-to-noise ratio in db
+        :param max_snr_in_db: Maximum signal-to-noise ratio in db
+        :param min_SNR: Minimum signal-to-noise ratio (legacy)
+        :param max_SNR: Maximum signal-to-noise ratio (legacy)
         :param p: The probability of applying this transform
+        :snr_characteristics: Use legacy mode or not
```
iver56, line 209: Remove this line

iver56 (changes requested):

audiomentations/augmentations/transforms.py
atamazian: Done.

iver56 (approved):

iver56:
Thanks

PR #112, opened 2021-06-17
input samples dtype can be float32 or 64

shawnthu:


iver56:
As far as I know, 64 bits is a waste of storage/memory, so that's why I don't think it should be encouraged/allowed for normal audio applications. If you have a use case for 64 bits, can you please explain?

* CD quality uses a 16 bit format, which is good enough for most audio
* Audio professionals often record audio in a 24 bit format
* Professional DAWs designed to mix hundreds of channels of audio may use 32 bits as the internal bit depth

Maybe instead of allowing float64, we could convert it to float32 and show a warning that recommends using float32 for audio?

atamazian:
I agree that a conversion would be more optimal in this case.

shawnthu:
It's no need. Sorry.

shawnthu:
I made some mistakes. LOL

iver56:
Ok. I've converted this to an issue. Feel free to pick up the task if you want:

https://github.com/iver56/audiomentations/issues/115

Issue #9, opened 2019-02-18
Add band-pass filter transformation


iver56:

iver56:
https://pyaudiodsptools.readthedocs.io/en/latest/

atamazian:
Please assign me to this issue.

atamazian:
This issue is solved, please close it.

Issue #7, opened 2019-02-18
Add high pass filter transformation


iver56:

iver56:
https://github.com/jiaaro/pydub/blob/master/pydub/effects.py#L251

iver56:
https://pyaudiodsptools.readthedocs.io/en/latest/

iver56:
https://github.com/asteroid-team/torch-audiomentations#highpassfilter

atamazian:
Please assign me to this issue.

atamazian:
This issue is solved, please close it.

PR #114, opened 2021-06-19
Add low-pass, high-pass and band-pass filter transforms

atamazian:
Implementing https://github.com/iver56/audiomentations/issues/7, https://github.com/iver56/audiomentations/issues/8 and https://github.com/iver56/audiomentations/issues/9.

### DIFF
--- a/audiomentations/augmentations/transforms.py
+++ b/audiomentations/augmentations/transforms.py
@@ -1209,7 +1209,7 @@ class LowPassFilter(BaseWaveformTransform):
         seg = low_pass_filter(
             seg, cutoffs_as_fraction_of_sample_rate
         )
-        samples = seg.get_array_of_samples()
+        samples = np.array(seg.get_array_of_samples())
         return samples
 
 
--- a/audiomentations/augmentations/transforms.py
+++ b/audiomentations/augmentations/transforms.py
@@ -1194,12 +1194,12 @@ class LowPassFilter(BaseWaveformTransform):
         )
 
     def apply(self, samples: np.array, sample_rate: int = None):
-        num_samples = samples.shape[-1]
+        num_channels = samples.shape[0]
         seg = AudioSegment(
             samples.tobytes(), 
             frame_rate=sample_rate,
             sample_width=samples.dtype.itemsize, 
-            channels=num_samples
+            channels=num_channels
         )
 
         cutoffs_as_fraction_of_sample_rate = (
--- a/audiomentations/augmentations/transforms.py
+++ b/audiomentations/augmentations/transforms.py
@@ -1165,20 +1165,14 @@ class LowPassFilter(BaseWaveformTransform):
         self,
         min_cutoff_freq=150,
         max_cutoff_freq=7500,
-        mode: str = "per_example",
         p: float = 0.5,
-        p_mode: str = None,
-        sample_rate: int = None,
     ):
         """
         :param min_cutoff_freq: Minimum cutoff frequency in hertz
         :param max_cutoff_freq: Maximum cutoff frequency in hertz
-        :param mode:
         :param p:
-        :param p_mode:
-        :param sample_rate:
         """
-        super().__init__(mode, p, p_mode, sample_rate)
+        super().__init__(p)
 
         self.min_cutoff_freq = min_cutoff_freq
         self.max_cutoff_freq = max_cutoff_freq
@@ -1230,20 +1224,14 @@ class HighPassFilter(BaseWaveformTransform):
         self,
         min_cutoff_freq=20,
         max_cutoff_freq=2400,
-        mode: str = "per_example",
-        p: float = 0.5,
-        p_mode: str = None,
-        sample_rate: int = None,
+        p: float = 0.5
     ):
         """
         :param min_cutoff_freq: Minimum cutoff frequency in hertz
         :param max_cutoff_freq: Maximum cutoff frequency in hertz
-        :param mode:
         :param p:
-        :param p_mode:
-        :param sample_rate:
         """
-        super().__init__(mode, p, p_mode, sample_rate)
+        super().__init__(p)
 
         self.min_cutoff_freq = min_cutoff_freq
         self.max_cutoff_freq = max_cutoff_freq
@@ -1297,22 +1285,16 @@ class BandPassFilter(BaseWaveformTransform):
         max_low_cutoff_freq=7500,
         min_high_cutoff_freq=15000,
         max_high_cutoff_freq=22350,
-        mode: str = "per_example",
-        p: float = 0.5,
-        p_mode: str = None,
-        sample_rate: int = None,
+        p: float = 0.5
     ):
         """
         :param min_low_cutoff_freq: Minimum low cutoff frequency in hertz
         :param max_low_cutoff_freq: Maximum low cutoff frequency in hertz
         :param min_high_cutoff_freq: Minimum high cutoff frequency in hertz
         :param max_high_cutoff_freq: Maximum high cutoff frequency in hertz
-        :param mode:
         :param p:
-        :param p_mode:
-        :param sample_rate:
         """
-        super().__init__(mode, p, p_mode, sample_rate)
+        super().__init__(p)
 
         self.min_low_cutoff_freq = min_low_cutoff_freq
         self.max_low_cutoff_freq = max_low_cutoff_freq
--- a/audiomentations/augmentations/transforms.py
+++ b/audiomentations/augmentations/transforms.py
@@ -1170,7 +1170,7 @@ class LowPassFilter(BaseWaveformTransform):
         """
         :param min_cutoff_freq: Minimum cutoff frequency in hertz
         :param max_cutoff_freq: Maximum cutoff frequency in hertz
-        :param p:
+        :param p: The probability of applying this transform
         """
         super().__init__(p)
 
@@ -1228,7 +1228,7 @@ class HighPassFilter(BaseWaveformTransform):
         """
         :param min_cutoff_freq: Minimum cutoff frequency in hertz
         :param max_cutoff_freq: Maximum cutoff frequency in hertz
-        :param p:
+        :param p: The probability of applying this transform
         """
         super().__init__(p)
 
@@ -1279,29 +1279,29 @@ class BandPassFilter(BaseWaveformTransform):
 
     def __init__(
         self,
-        min_low_cutoff_freq=150,
-        max_low_cutoff_freq=7500,
-        min_high_cutoff_freq=15000,
-        max_high_cutoff_freq=22350,
-        p: float = 0.5
+        min_center_freq=100.0,
+        max_center_freq=1000.0,
+        min_q=1.0,
+        max_q=2.0,
+        p=0.5
     ):
         """
-        :param min_low_cutoff_freq: Minimum low cutoff frequency in hertz
-        :param max_low_cutoff_freq: Maximum low cutoff frequency in hertz
-        :param min_high_cutoff_freq: Minimum high cutoff frequency in hertz
-        :param max_high_cutoff_freq: Maximum high cutoff frequency in hertz
-        :param p:
+        :param min_center_freq: Minimum center frequency in hertz
+        :param max_center_freq: Maximum center frequency in hertz
+        :param min_q: Minimum ratio of center frequency to bandwidth
+        :param max_q: Maximum ratio of center frequency to bandwidth
+        :param p: The probability of applying this transform
         """
         super().__init__(p)
 
-        self.min_low_cutoff_freq = min_low_cutoff_freq
-        self.max_low_cutoff_freq = max_low_cutoff_freq
-        self.min_high_cutoff_freq = min_high_cutoff_freq
-        self.max_high_cutoff_freq = max_high_cutoff_freq
-        if self.min_low_cutoff_freq > self.max_low_cutoff_freq:
-            raise ValueError("min_low_cutoff_freq must not be greater than max_low_cutoff_freq")
-        if self.min_high_cutoff_freq > self.max_high_cutoff_freq:
-            raise ValueError("min_high_cutoff_freq must not be greater than max_high_cutoff_freq")
+        self.min_center_freq = min_center_freq
+        self.max_center_freq = max_center_freq
+        self.min_q = min_q
+        self.max_q = max_q
+        if self.min_center_freq > self.max_center_freq:
+            raise ValueError("min_center_freq must not be greater than max_center_freq")
+        if self.min_q > self.max_q:
+            raise ValueError("min_q must not be greater than max_q")
 
     def randomize_parameters(
         self, selected_samples: np.array, sample_rate: int = None
@@ -1311,14 +1311,14 @@ class BandPassFilter(BaseWaveformTransform):
         """
         batch_size, _, num_samples = selected_samples.shape
 
-        self.parameters["low_cutoff_freq"] = np.random.uniform(
-            low=self.min_low_cutoff_freq,
-            high=self.max_low_cutoff_freq,
+        self.parameters["center_freq"] = np.random.uniform(
+            low=self.min_center_freq,
+            high=self.max_center_freq,
             size=batch_size
         )
-        self.parameters["high_cutoff_freq"] = np.random.uniform(
-            low=self.min_high_cutoff_freq,
-            high=self.max_high_cutoff_freq,
+        self.parameters["q"] = np.random.uniform(
+            low=self.min_q,
+            high=self.max_q,
             size=batch_size
         )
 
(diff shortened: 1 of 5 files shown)


audiomentations/augmentations/transforms.py
```diff
@@ -1151,7 +1152,200 @@ def apply(self, samples, sample_rate):
...
+        # TODO: Instead of using a for loop, perform batched compute to speed things up
+        for i in range(batch_size):
+            selected_samples[i] = high_pass_filter(
+                selected_samples[i], cutoffs_as_fraction_of_sample_rate[i].item()
+            )
+
+        return selected_samples   
+
+
+class BandPassFilter(BaseWaveformTransform):
```
iver56, line 1272: I would feel more comfortable having two parameters center frequency and bandwidth instead of two parameters HPF cutoff and LPF cutoff. That also eliminates any potential issues where low cutoff > high cutoff, leading to a silent output.

It probably makes sense to have the bandwidth be related to the chosen center frequency. Maybe the bandwidth can be a float that denotes the "ratio of the center frequency"?

Some bandpass filter developers like to have a bandwidth parameter with inverse behavior, and they call it Q. An example of that can be found here: https://github.com/facebookresearch/AugLy/blob/main/augly/audio/transforms.py#L561

iver56:
Thanks for the initiative :) The LPF and HPF look good. If you want to improve the frequency selector somewhat, you can uniformly pick cutoff frequencies in the mel scale instead of in hz. The motivation behind this is that human perception of pitch does not linearly relate to hz. Not sure how relevant that is for data augmentation though.

iver56:
https://en.wikipedia.org/wiki/Mel_scale

audiomentations/augmentations/transforms.py
atamazian: Which parameters should be selected randomly in that case? Center_freq, Q, or both of them?

iver56: Center freq and Q could both be selected randomly

iver56: People who want e.g. a constant Q can set min Q and max Q to the same value :)

atamazian: OK, got it.

atamazian:
I'm not sure if mel scale is needed here.

audiomentations/augmentations/transforms.py
atamazian: Done!

iver56:
Thanks for your patience. I will re-review this soon-ish. I might want to wrap up the upcoming release of 0.17.0 first, and then merge this afterwards. I haven't decided yet.

audiomentations/augmentations/transforms.py
```diff
@@ -1151,7 +1152,200 @@ def apply(self, samples, sample_rate):
...
+        batch_size, _, num_samples = selected_samples.shape
+
+        self.parameters["cutoff_freq"] = np.random.uniform(
+            low=self.min_cutoff_freq,
+            high=self.max_cutoff_freq,
+            size=batch_size
+        )
+
+    def apply_transform(self, selected_samples: np.array, sample_rate: int = None):
+        batch_size, num_channels, num_samples = selected_samples.shape
```
iver56, line 1197: Audiomentations only supports 1D ndarrays (mono audio) and 2D ndarrays (for stereo/multichannel audio). Can you please remove the assumption of the existence of a batch dimension? It should work on both 1D and 2D audio, but not 3D audio :)

audiomentations/augmentations/transforms.py
```diff
@@ -1151,7 +1152,200 @@ def apply(self, samples, sample_rate):
...
+        """
+        batch_size, _, num_samples = selected_samples.shape
+
+        self.parameters["cutoff_freq"] = np.random.uniform(
+            low=self.min_cutoff_freq,
+            high=self.max_cutoff_freq,
+            size=batch_size
+        )
+
+    def apply_transform(self, selected_samples: np.array, sample_rate: int = None):
```
iver56, line 1196: To be consistent with other transforms, please call the parameter `samples` instead of `selected_samples`

iver56 (changes requested):

audiomentations/augmentations/transforms.py
atamazian: I've fixed that.

audiomentations/augmentations/transforms.py
atamazian: My fault:) I was using `torch-audiomentations` code as a reference, that's why this mistake appeared. I've fixed the code.

audiomentations/augmentations/transforms.py
```diff
@@ -1151,7 +1152,193 @@ def apply(self, samples, sample_rate):
...
+            low=self.min_center_freq,
+            high=self.max_center_freq
+        )
+        self.parameters["q"] = np.random.uniform(
+            low=self.min_q,
+            high=self.max_q
+        )
+
+    def apply_transform(self, samples: np.array, sample_rate: int = None):
+        num_channels, num_samples = samples.shape
```
iver56, line 1320: This will break for 1D numpy arrays and only work for 2D numpy arrays. It looks like you are not using num_channels here - perhaps you can do this instead:

num_samples = samples.shape[-1]

iver56 (changes requested):

audiomentations/augmentations/transforms.py
atamazian: I've fixed that.

iver56:
Thanks for the update. I'm on vacation, so my responses can be on the slow side these days. Thank you for your patience!

audiomentations/augmentations/transforms.py
```diff
@@ -1151,7 +1152,193 @@ def apply(self, samples, sample_rate):
...
+
+        low_cutoffs_as_fraction_of_sample_rate = (
+            self.parameters["center_freq"] * (1 - 0.5 / self.parameters["q"])  / sample_rate
+        )
+        high_cutoffs_as_fraction_of_sample_rate = (
+            self.parameters["center_freq"] * (1 + 0.5 / self.parameters["q"])  / sample_rate
+        )
 
+        samples = low_pass_filter(
+            samples, low_cutoffs_as_fraction_of_sample_rat.item()
```
iver56, line 1333: Unresolved reference `low_cutoffs_as_fraction_of_sample_rat`

This error was not caught by any of the unit tests. I would rest a bit easier if this was covered by a test

audiomentations/augmentations/transforms.py
```diff
@@ -1151,7 +1152,193 @@ def apply(self, samples, sample_rate):
...
+        self.parameters["center_freq"] = np.random.uniform(
+            low=self.min_center_freq,
+            high=self.max_center_freq
+        )
+        self.parameters["q"] = np.random.uniform(
+            low=self.min_q,
+            high=self.max_q
+        )
+
+    def apply_transform(self, samples: np.array, sample_rate: int = None):
```
iver56, line 1319: This must be called `apply` instead of `apply_transform`, or else this function won't be run by `BaseWaveformTransform`. Until this is fixed this transform seems to be accidentally a no-op 😅

audiomentations/augmentations/transforms.py
```diff
@@ -1151,7 +1152,193 @@ def apply(self, samples, sample_rate):
...
+        :params samples: (num_channels, num_samples)
+        """
+        num_samples = samples.shape[-1]
+
+        self.parameters["cutoff_freq"] = np.random.uniform(
+            low=self.min_cutoff_freq,
+            high=self.max_cutoff_freq
+        )
+
+    def apply_transform(self, samples: np.array, sample_rate: int = None):
```
iver56, line 1251: Please rename this to apply

audiomentations/augmentations/transforms.py
```diff
@@ -1151,7 +1152,193 @@ def apply(self, samples, sample_rate):
...
```
iver56, line 1195: Please rename this to apply

tests/test_band_pass_filter.py
```diff
@@ -0,0 +1,29 @@
...
+        augment = BandPassFilter(min_center_freq=100, 
+                                 max_center_freq=5000,
+                                 min_q=1.0,
+                                 max_q=2.0,
+                                 p=1.0)
+        processed_samples = augment(
+            samples=samples, sample_rate=sample_rate
+        )
+        self.assertEqual(processed_samples.shape, samples.shape)
+        self.assertEqual(processed_samples.dtype, np.float32)
```
iver56, line 29: It could be useful to check that `processed_samples` is not equal to `samples` here

https://stackoverflow.com/questions/38506044/numpy-testing-assert-array-not-equal

iver56 (changes requested):

audiomentations/augmentations/transforms.py
atamazian: Sorry for such lame mistakes) I've fixed this.

audiomentations/augmentations/transforms.py
atamazian: Fixed.

audiomentations/augmentations/transforms.py
atamazian: Fixed.

audiomentations/augmentations/transforms.py
atamazian: Fixed. As for test, I can't understand myself why it wasn't caught. Do you have any idea about it? `cutoff` argument is obligatory in `low_pass_filter`, there's no way to bypass it.

tests/test_band_pass_filter.py
atamazian: But it's possible to have `processed_samples` equal to `samples` if signal fits to bandwidth of the filter.

audiomentations/augmentations/transforms.py
iver56: I think it wasn't caught because the function did not run at all (due to the function name). It's probably okay now

atamazian: OK

tests/test_band_pass_filter.py
iver56: Yes, but that can be solved by setting a seed for the random generator in the beginning of the test, so that the same parameters are used every time the test is run.

The intention of this extra assertion would be to check that the transform wasn't accidentally a no-op :)

audiomentations/augmentations/transforms.py
```diff
@@ -1151,7 +1152,193 @@ def apply(self, samples, sample_rate):
...
+        self.parameters["cutoff_freq"] = np.random.uniform(
+            low=self.min_cutoff_freq,
+            high=self.max_cutoff_freq
+        )
+
+    def apply(self, samples: np.array, sample_rate: int = None):
+        num_samples = samples.shape[-1]
+
+        if sample_rate is None:
+            sample_rate = self.sample_rate
```
iver56, line 1199: The class has no self.sample_rate, so this line and the one above can be removed

audiomentations/augmentations/transforms.py
```diff
@@ -1151,7 +1152,193 @@ def apply(self, samples, sample_rate):
...
+        if self.min_cutoff_freq > self.max_cutoff_freq:
+            raise ValueError("min_cutoff_freq must not be greater than max_cutoff_freq")
+
+    def randomize_parameters(
+        self, samples: np.array, sample_rate: int = None
+    ):
+        """
+        :params samples: (num_channels, num_samples)
+        """
+        num_samples = samples.shape[-1]
```
iver56, line 1188: This function needs a call like `super().randomize_parameters(samples, sample_rate)` in the beginning to make sure that `self.parameters["should_apply"]` is set correctly. Otherwise `apply` won't be run :/

iver56 (changes requested):

audiomentations/augmentations/transforms.py
atamazian: Fixed.

audiomentations/augmentations/transforms.py
atamazian: Fixed.

audiomentations/augmentations/transforms.py
```diff
@@ -1151,7 +1152,191 @@ def apply(self, samples, sample_rate):
...
```
iver56, line 1242:
```suggestion
        super().randomize_parameters(samples, sample_rate)
```

audiomentations/augmentations/transforms.py
```diff
@@ -1151,7 +1152,191 @@ def apply(self, samples, sample_rate):
...
+        if self.min_q > self.max_q:
+            raise ValueError("min_q must not be greater than max_q")
+
+    def randomize_parameters(
+        self, samples: np.array, sample_rate: int = None
+    ):
+        """
+        :params samples: (num_channels, num_samples)
+        """
+        num_samples = samples.shape[-1]
```
iver56, line 1306:
```suggestion
        super().randomize_parameters(samples, sample_rate)
```

iver56 (changes requested):

audiomentations/augmentations/transforms.py
```diff
@@ -1151,7 +1152,191 @@ def apply(self, samples, sample_rate):
...
+
+    def apply(self, samples: np.array, sample_rate: int = None):
+        num_samples = samples.shape[-1]
+
+        cutoffs_as_fraction_of_sample_rate = (
+            self.parameters["cutoff_freq"] / sample_rate
+        )
+
+        samples = low_pass_filter(
+            samples, cutoffs_as_fraction_of_sample_rate.item()
```
iver56, line 1204: When I run the tests, I get `E       AttributeError: 'float' object has no attribute 'item'`

(this goes for all the three transforms).

I think we need to remove .item() here

iver56 (changes requested):

atamazian:
I've encountered some more error, so I'll need some time to fix them.

atamazian:
I've fixed all error (had to rewrite some stuff), please check. I also changed test files - I added plots for easy visual control.

audiomentations/augmentations/transforms.py
atamazian: Code was changed.

tests/test_band_pass_filter.py
atamazian: No-op check can be conducted now by checking output plots.

iver56:
Thanks, I'll have a look soon-ish (maybe today or tomorrow)

tests/test_band_pass_filter.py
```diff
@@ -0,0 +1,32 @@
...
+        processed_samples = augment(
+            samples=samples, sample_rate=sample_rate
+        )
+
+        self.assertEqual(processed_samples.shape, samples.shape)
+        self.assertEqual(processed_samples.dtype, np.float32)
+        
+        plt.plot(samples)
+        plt.plot(processed_samples, '-.')
+        plt.show()
```
iver56, line 32: Please wrap plotting code in tests inside something like an `if DEBUG:` block. You can put the DEBUG constant (default to False) below the imports of the file. When needed, one can turn this on manually and look at the plots.

Motivation: The test runner (e.g. the one on Github) may get stuck while trying to run the test suite if it should show plots while running the tests

audiomentations/augmentations/transforms.py
```diff
@@ -1151,7 +1154,236 @@ def apply(self, samples, sample_rate):
...
+        samples = convert_int16_samples_to_float(np.array(audio_segment.get_array_of_samples()))
+        return samples
+
+
+class BandPassFilter(BaseWaveformTransform):
+    """
+    Apply band-pass filtering to the input audio.
+    """
+
+    supports_multichannel = True
```
iver56, line 1305: This must be set to False to be aligned with the implementation

audiomentations/augmentations/transforms.py
```diff
@@ -1151,7 +1154,236 @@ def apply(self, samples, sample_rate):
...
+            print(
+                "Failed to import pydub. Maybe it is not installed? "
+                "To install the optional pydub dependency of audiomentations,"
+                " do `pip install audiomentations[extras]` instead of"
+                " `pip install audiomentations`",
+                file=sys.stderr,
+            )
+            raise
+            
+        assert len(samples.shape) == 1
```
iver56, line 1364: This check is not needed here, as the base class (BaseWaveformTransform) already runs an `is_multichannel` check and throws an exception if multichannel audio is given to a mono-only transform

iver56 (changes requested):

tests/test_band_pass_filter.py
atamazian: Done

audiomentations/augmentations/transforms.py
atamazian: Fixed.

audiomentations/augmentations/transforms.py
atamazian: Fixed.

iver56:
Thanks for the contribution :) I think I'll merge this now and leave some nice-to-haves for future work:

* Add support for multichannel audio
* Check other implementations to see if they are faster than what pydub provides
* Add support for steeper filters - 6 dB per octave is quite gentle. We might want to support 12, 18 and 24 dB as well.
* More comprehensive unit test suite, e.g. test both single-channel and multichannel audio, and automatically check that the resulting spectrum is as expected

iver56 (approved):

Issue #115, opened 2021-06-22
If input is float64, convert to float32 and show a warning


iver56:
For context, see https://github.com/iver56/audiomentations/pull/112#issuecomment-863555050

atamazian:
Please assign this to me.

atamazian:
Please close this issue since it's fixed.

PR #118, opened 2021-06-29
If input is float64, convert to float32 and show a warning

atamazian:
Implementing https://github.com/iver56/audiomentations/issues/115


iver56:
I think this should be done in the __call__ function of the base class instead of only Mp3Compression.apply_lameenc

https://github.com/iver56/audiomentations/blob/master/audiomentations/core/transforms_interface.py#L53

atamazian:
Done

iver56:
Thanks

Issue #117, opened 2021-06-28
Reverse audio ("time inversion")


iver56:
Reversing the audio can be relevant sometimes.

Inspired by this paper: https://arxiv.org/pdf/2106.13043.pdf

atamazian:
Please assign me to this task.

atamazian:
Please close this issue.

PR #119, opened 2021-07-02
Add Reverse audio

atamazian:
Adds new augmentation called `Reverse`, which allows to reverse audio. Implements https://github.com/iver56/audiomentations/issues/117.

### DIFF
--- a/tests/test_reverse.py
+++ b/tests/test_reverse.py
@@ -1,0 +1,27 @@
+import unittest
+
+import numpy as np
+
+from audiomentations import Reverse
+
+
+class TestReverse(unittest.TestCase):
+    def test_single_channel(self):
+        samples = np.array([0.5, 0.6, -0.2, 0.0], dtype=np.float32)
+        sample_rate = 16000
+        augmenter = Reverse(p=1.0)
+        samples = augmenter(samples=samples, sample_rate=sample_rate)
+
+        self.assertEqual(samples.dtype, np.float32)
+        self.assertEqual(len(samples), 4)
+
+    def test_multichannel(self):
+        samples = np.array(
+            [[0.9, 0.5, -0.25, -0.125, 0.0], [0.95, 0.5, -0.25, -0.125, 0.0]],
+            dtype=np.float32,
+        )
+        sample_rate = 16000
+        augmenter = Reverse(p=1.0)
+        samples = augmenter(samples=samples, sample_rate=sample_rate)
+
+        self.assertEqual(samples.dtype, np.float32)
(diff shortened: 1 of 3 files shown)


tests/test_reverse.py
```diff
@@ -0,0 +1,27 @@
+import unittest
+
+import numpy as np
+
+from audiomentations import Reverse
```
iver56, line 5: When running `pytest`, I get the following error:

```______________________________________ ERROR collecting tests/test_reverse.py ______________________________________
ImportError while importing test module 'C:\Users\Iver\Code\audiomentations\tests\test_reverse.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests\test_reverse.py:5: in <module>
    from audiomentations import Reverse
E   ImportError: cannot import name 'Reverse' from 'audiomentations' (C:\Users\Iver\Code\audiomentations\audiomentations\__init__.py)
```

To fix this error, either change the import here or add `Reverse` to the list of publicly exposed transforms in audiomentations/\_\_init\_\_.py

tests/test_reverse.py
```diff
@@ -0,0 +1,27 @@
...
+    def test_multichannel(self):
+        samples = np.array(
+            [[0.9, 0.5, -0.25, -0.125, 0.0], [0.95, 0.5, -0.25, -0.125, 0.0]],
+            dtype=np.float32,
+        )
+        sample_rate = 16000
+        augmenter = Reverse(p=1.0)
+        samples = augmenter(samples=samples, sample_rate=sample_rate)
+
+        self.assertEqual(samples.dtype, np.float32)
```
iver56, line 27: Please add an assertion that checks that the output is reversed as expected when applied to multichannel audio:

```
        assert_array_almost_equal(
            reversed_samples,
            np.array(
                [[0.0, -0.125, -0.25, -0.5, 0.9], [0.0, -0.125, -0.25, 0.5, 0.95]],
                dtype=np.float32,
            ),
        )
```

You may find that the test fails after you add this assertion. In that case, the implementation needs to be fixed. Maybe slice notation is better than np.flipud?

iver56 (changes requested):
Thanks for the initiative :)

tests/test_reverse.py
atamazian: Done.

tests/test_reverse.py
atamazian: Can you tell me what is the source of `assert_array_almost_equal` function?

iver56: Sorry, I forgot to mention that. It's `from numpy.testing import assert_array_almost_equal`

atamazian: I've fixed that problem and updated test file, please check.

iver56 (approved):

iver56:
Thanks!

Issue #57, opened 2020-09-24
Add a wow and flutter transform


iver56:
Inspired by https://arxiv.org/ftp/arxiv/papers/1912/1912.05472.pdf

![bilde](https://user-images.githubusercontent.com/1470603/94115994-c2f39600-fe4a-11ea-820e-425f1f3dca7c.png)

iver56:
Maybe this can be used? https://github.com/sevagh/audio-degradation-toolbox

atamazian:
Please assign me to this issue.

iver56:
Thanks for the update 👍 And thanks for giving it a try

PR #120, opened 2021-07-03
Add ApplyWowResampling

atamazian:
Implementing https://github.com/iver56/audiomentations/issues/57.

### DIFF
--- a/audiomentations/__init__.py
+++ b/audiomentations/__init__.py
@@ -5,6 +5,7 @@ from .augmentations.transforms import (
     AddGaussianSNR,
     AddImpulseResponse,
     ApplyImpulseResponse,
+    ApplyWowResampling,
     AddShortNoises,
     Clip,
     ClippingDistortion,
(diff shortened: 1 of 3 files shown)


iver56:
I'm on vacation, so my time is limited these days. I'll get to this eventually. Meanwhile, thanks for your patience :)

iver56:
The implementation here seems to do something else than what is intended. Wow and flutter is a change of pitch in the audio over time. For an example, listen to [this youtube video](https://www.youtube.com/watch?v=kCwRdrFtJuE) from 0:37. The implementation in this pull request mixes in a distorted variant of the original signal, which is not the same thing as wow and flutter.

I can see how you were confused by the referenced paper, which states

bilde

However, the `x` mentioned here refers to a position (i.e. time) in the audio file, not the array of audio samples.

In other words, that function maps a position in the original audio to a position in the output audio.

To be able to sample smoothly from different positions in the audio, you would have to do some interpolation (e.g. sinc) to make it sound good.

atamazian:
I've got the idea.

iver56:
Do you plan to update this pull request? Or should I close it for now?

atamazian:
Close it for now. Maybe later I'll finish this.

Issue #88, opened 2021-02-20
Random interruption pulses to augment ECG lead-off condition


darkcurrent:
Hello,

I just wrote a function, for ECG augmentation. The main purpose is simulate the signals when the ECG electrodes are in lead-off condition. It can also be used for audio data augmentation.
The function generates random number of flat pulses with some noise on it. The pulses' amplitudes are between signal max-min range and pulses have random width. 
Maybe the function can be added here.

Regards,

```python
import wavio
import numpy as np
from matplotlib import pyplot as plt

audio_seg=wavio.read("sample_audio.wav").data.astype(np.float32).T[0]

def InterruptSignal(audio_seg, max_num_interruptions, noise_level, p):
    # audio_seg:             Audio segment (1D Numpy array)
    # max_num_interruptions: Maximum number of interruption pulses. Pulses are generated randomly
    # noise_level:           The level of noise to be added the pulses. "CNN's don't like constant values :)"
    # p:                     Triggering probability
    if np.random.random() < p / 1.:
        max_peak = np.max(audio_seg).astype(int)
        min_peak = np.min(audio_seg).astype(int)
        len_audio                = audio_seg.shape[-1]
        num_random_interruptions = np.random.randint(max_num_interruptions)+1
        interruption_start_times = (len_audio * np.random.random(num_random_interruptions)*4/5).astype(int)
        for interruption in interruption_start_times:
            interruption_len = int(len_audio*np.random.random()//5)
            interruption_val = np.random.randint(min_peak, max_peak)
            audio_seg[interruption:interruption+interruption_len] = interruption_val+np.random.normal(0, 1, interruption_len)*noise_level
    return audio_seg

fig, (ax1, ax2) = plt.subplots(2)
fig.suptitle('Two random interruptions with some noise')
ax1.plot(audio_seg[:5000])
ax2.plot(InterruptSignal(audio_seg[:5000],
                         max_num_interruptions=3,
                         noise_level=100,
                         p=1))
plt.show()
```
<img width="294" alt="interruptions" src="https://user-images.githubusercontent.com/6089944/108605641-75159680-73c6-11eb-950f-ffa38f1d93fb.PNG">

iver56:
Interesting 👍 It can be added. Or maybe it can be an option in AddShortNoises?

darkcurrent:
Hmm they looks so similar...
I think the only difference is the DC offset. So making an option to AddShortNoises makes more sense.

atamazian:
[...]

PR #122, opened 2021-07-20
Add InterruptPulse

atamazian:
Attempt to implement https://github.com/iver56/audiomentations/issues/88.


audiomentations/augmentations/transforms.py
```diff
@@ -1241,6 +1241,56 @@ def apply(self, samples, sample_rate):
...
+    ):
+        super().randomize_parameters(samples, sample_rate)
+
+        self.parameters["num_interruptions"] = np.random.randint(
+            low=self.min_num_interruptions,
+            high=self.max_num_interruptions
+        )
+        
+    def apply(self, samples: np.array, sample_rate: int = None):
+        max_peak = np.max(samples).astype(int)
```
iver56, line 1282: This seems a bit off, as the incoming samples is expected to be audio with float32 samples between -1.0 and 1.0. Unless the audio is maxed out in some places, max_peak and min_peak will be zero.

Did you assume the input to be int16 with max/min values like +/- 32767?

iver56 (changes requested):
Thanks for the contribution :)

I briefly looked at the code, and have noted some initial thoughts below:

* I would like interruption start times, lenghts and values to be selected and remembered in `randomize_parameters`. The idea is that one can run the augmentation, freeze the parameters, then run the same transform (with the same params/perturbations) on another audio with similar results. E.g. it could be useful for applying the same transform to two recordings (from two microphones in the same room recording at the same time). It could also be useful for perturbing an input and a target in the same way.
* Maybe dB is a better unit for the noise level?
* A default noise level of 100 seems too high, given that input audio will have samples between -1 and 1. I have not tried to run the code though.
* I noticed that some lines were too long. Could you run `black` on the code, please? See also https://github.com/psf/black

@darkcurrent Do you have any comments regarding this pull request?

audiomentations/augmentations/transforms.py
```diff
@@ -1241,6 +1241,56 @@ def apply(self, samples, sample_rate):
         return distorted_samples
 
     
+class InterruptPulse(BaseWaveformTransform):
+    """
+    Add interrupt pulses to the input audio.
```
iver56, line 1246: We should mention here that this transform could be useful for ECG augmentation. Further, we can use some of the description from @darkcurrent:

> The main purpose is simulate the signals when the ECG electrodes are in lead-off condition. It can also be used for audio data augmentation.
The function generates random number of flat pulses with some noise on it. The pulses' amplitudes are between signal max-min range and pulses have random width.

audiomentations/augmentations/transforms.py
atamazian: Yes, I've assumed so. I've added conversion to int in the beginning, and to float in the end. Or maybe it's better to change code to work with float?

audiomentations/augmentations/transforms.py
atamazian: Added.

audiomentations/augmentations/transforms.py
iver56: In my opinion, it's better to change the code so it works with float directly :) That way, there is no need for the conversion

atamazian: Hmm, OK. I'll try.

iver56:
Hi atamazian :) What's the status on this?

atamazian:
I'm still working on this... Sorry for the delay.

iver56:
Thanks for the update! I'll have a look

By the way, I recently changed the code structure (because transforms.py became too large), so there's a merge conflict in this PR now. You could fix the merge conflict by adding your transform in a new file interrupt_pulse.py instead of adding it to transforms.py

atamazian:
OK, I'll do it.

atamazian:
Is it OK now?

iver56:
It says there are still conflicts that must be resolved:

bilde

It it is easier for you, you could create a new branch (from master) and create a new PR

atamazian:
Conflicts resolved!

iver56:
I added this transform to the demo script, and identified some issues:

* `from .augmentations.interrupt_pulse import InterruptPulse` needs to be added in `audiomentations/__init__.py`
* The first audio clip seems to have been unaffected
* The second audio clip was perturbed, but the added noise is way too loud
* It looks like the transform alters the original input, while it should ideally leave it untouched

bilde

atamazian:
Got it, I'll see what can I do

atamazian:
Unfortunately I'm also not an expert in ECG sphere either.

iver56:
Yeah, I don't blame you 😄 You are trying to help, and I appreciate that.

@darkcurrent Is there any literature that supports the usefulness of a transform like this as a form of data augmentation? If there is, I would be more comfortable spending time on it. If not, I'm not sure if I'll want to spend the time to make this ready for release

iver56:
Thanks for the effort, but I'm closing this now in favor of https://github.com/iver56/audiomentations/issues/160

It will be a more general solution

PR #124, opened 2021-07-28
fix info about facebookresearch/WavAugment

EnisBerk:
Hey,
Thanks for the great library and info about other ones.

facebookresearch/WavAugment does not have GPU support yet. You can see the line giving the warning [there](https://github.com/facebookresearch/WavAugment/blob/54afcdb00ccc852c2f030f239f8532c9562b550e/augment/effects.py#L139)

The same info is in the limitations section of [this](https://github.com/facebookresearch/WavAugment/blob/master/examples/python/WavAugment_walkthrough.ipynb) notebook.


iver56:
Ah, you're right! Thanks for the correction.

PR #130, opened 2021-09-27
Changed the transform AddShortNoise. 

Cangonin:
An option to choose how the noises will be mixed in the input audio has been added. Two options exist now, "relative" and "absolute". If the chosen option is "relative", the power of the added noise will be proportional to the power of
 the input sound (as it was before by default). If the chosen option is "absolute", the added noises will all have the same power independently of the power of the input.


audiomentations/augmentations/transforms.py
```diff
@@ -1074,7 +1086,7 @@ def apply(self, samples, sample_rate):
             noise_gain[-fade_out_mask.shape[0] :] = np.minimum(
                 noise_gain[-fade_out_mask.shape[0] :], fade_out_mask
             )
-            noise_samples = noise_samples * noise_gain
+            noise_samples = noise_samples * noise_gain  # Gain here describes just the gain from the fade in and fade out.
```
iver56, line 1089: Line is too long. I recommend putting the comment above.

I also recommend checking out the python formatting tool called `black`, which we use in this project. I can help you set this up in your IDE 😄 

Black recommends a max line length of 90-ish characters.

audiomentations/augmentations/transforms.py
```diff
@@ -888,6 +888,8 @@ def __init__(
         max_snr_in_db=24,
         min_time_between_sounds=4.0,
         max_time_between_sounds=16.0,
+        noise_power=("relative", "absolute"),
+        absolute_noise_rms_db=-20,
```
iver56, line 892: In real life noises will vary in loudness. It would be nice to have two parameters that specify the min and max loudness of the noises to be mixed in. This way the augmentation transform can vary how loud the noises are, e.g. so it could mix in both soft and loud noises in the same audio clip.

audiomentations/augmentations/transforms.py
```diff
@@ -1095,16 +1107,30 @@ def apply(self, samples, sample_rate):
...
+                    noise_samples = noise_samples * gain
 
-                # Adjust the noise to match the desired noise RMS
-                noise_samples = noise_samples * (desired_noise_rms / noise_rms)
+                    max_noise_samples = np.amax(np.abs(noise_samples))
 
-                noise_placeholder[start_sample_index:end_sample_index] += noise_samples
+                    if max_noise_samples > 1:  # Not optimal, since this will change the rms. Happens primarily to
+                        # noise bursts.
+                        noise_samples = noise_samples / max_noise_samples
```
iver56, line 1131: Why do we have this peak normalization here? I think I would personally not add this feature, because as you say it changes the balance between signal and noise so that it doesn't match the snr provided given to \_\_init\_\_.

audiomentations already has several other transforms that can return results with values outside the [-1, 1] range, and I'm not too worried about it, because it can be dealt with at the end of the data augmentation pipeline. Values outside the [-1, 1] range is not really a problem when using float32, but it could result in clipping/distortion when converting it to int16 without normalizing it forst.

audiomentations/augmentations/transforms.py
```diff
@@ -907,6 +909,13 @@ def __init__(
             sounds/noises will be louder.
         :param min_time_between_sounds: Minimum pause time between the added sounds/noises
         :param max_time_between_sounds: Maximum pause time between the added sounds/noises
+        :param noise_power: Defines how the noises will be added to the audio input. If the chosen
+        option is "relative", the power of the added noise will be proportional to the power of
+        the input sound. If the chosen option is "absolute", the added noises will all have the
+        same power independently of the power of the input.
+        :param absolute_noise_rms_db: Is only used if noise_power is set to "absolute". The rms in dB
+         of the added noises will be constant and equal to absolute_noise_rms_db. Recommended values
+        for this parameter range from -10 to -50 (or even less).
```
iver56, line 918:
```suggestion
        :param noise_power: Defines how the noises will be added to the audio input. If the chosen
            option is "relative", the power of the added noise will be proportional to the power of
            the input sound. If the chosen option is "absolute", the added noises will all have the
            same power independently of the power of the input.
        :param absolute_noise_rms_db: Is only used if noise_power is set to "absolute". The rms in dB
            of the added noises will be constant and equal to absolute_noise_rms_db. Recommended values
            for this parameter range from -10 to -50 (or even less).
```

iver56 (changes requested):
It would be nice if you could add a unit test for this new feature. See https://github.com/iver56/audiomentations/blob/master/tests/test_add_short_noises.py

audiomentations/augmentations/transforms.py
```diff
@@ -888,6 +888,8 @@ def __init__(
         max_snr_in_db=24,
         min_time_between_sounds=4.0,
         max_time_between_sounds=16.0,
+        noise_power=("relative", "absolute"),
```
iver56, line 891: The default here should be "relative"

audiomentations/augmentations/transforms.py
Cangonin: I coded it this way because I was in the mindset of adding noises with the same power. But what you said is more flexible indeed, I'll fix it :)

audiomentations/augmentations/transforms.py
Cangonin: I added it because some handling noises were repeated quick bursts of sound with silence between the short peaks. This gave an artificially low noise rms and the max value of the peaks after treatment could reach 4,5. I was afraid that the initial speech recording would be hidden because of this. I could maybe calculate the rms of the noises another way, with audio event detection for example

iver56: I see. Hmm. How about:

1) Divide the short noise into short chunks (maybe 25 ms each? I dunno)
2) Calculate the RMS for each chunk
3) Order the RMS values and take the 90th percentile or something, and treat that as the overall RMS?

Just using my intuition here

iver56: Maybe it should be configurable with a parameter how we calculate the RMS?

For more or less stationary short noises mean RMS is okay, but for highly nonstationary noises, the more complicated percentile-based approach seems more useful

Cangonin: Yes, I was thinking about this approach too :)

demo/demo.py
```diff
@@ -132,6 +132,9 @@ def __exit__(self, type, value, traceback):
                 sounds_path=os.path.join(DEMO_DIR, "short_noises"),
                 min_snr_in_db=0,
                 max_snr_in_db=8,
+                min_absolute_noise_rms_db=-30,
+                max_absolute_noise_rms_db=-10,
+                noise_power="absolute",
```
iver56, line 137: Maybe it's better to have two separate configs of AddShortNoises in the demo script, one with noise_power="relative" and one with noise_power="absolute"? Then we can listen to both. I also did this with ApplyImpulseResponse a few lines above, and I gave the second one a different `"name"` 😃

Cangonin: Yes, good idea ! I'll change it :)

audiomentations/augmentations/transforms.py
```diff
@@ -1094,17 +1121,32 @@ def apply(self, samples, sample_rate):
                 end_sample_index = num_samples
 
             clean_rms = calculate_rms(samples[start_sample_index:end_sample_index])
-            noise_rms = calculate_rms(noise_samples)
+            noise_rms = calculate_rms_without_silence(noise_samples, sample_rate)
```
iver56, line 1124: This changes the default behavior of AddShortNoises.

Maybe we can have a parameter like `include_silence_in_noise_rms_estimation` (boolean) for choosing between `calculate_rms` and `calculate_rms_without_silence`?

The idea is that it would be nice to keep AddShortNoises backwards-compatible for existing users of audiomentations, so at this time include_silence_in_noise_rms_estimation should ideally default to True. Add a docstring that explains what the parameter does and when it is useful (when the noises are non-stationary and have silent parts in them). Then users who know what they are doing can set the value to False

iver56 (changes requested):
I had a listen to the output of the demo script, and it sounds good already 👍 

I just have two more comments now

iver56 (approved):

PR #137, opened 2021-11-04
Implement OneOf. Refactor compose classes. 

iver56:
Also add `apply_to_children` argument in some methods in BaseCompose. It defaults to `True`, so the change is backwards compatible.

#61

### DIFF
--- a/audiomentations/core/composition.py
+++ b/audiomentations/core/composition.py
@@ -113,8 +113,8 @@ class OneOf(BaseCompose):
 
     ```
     augment = OneOf([
-        TimeStretch(min_rate=0.8, max_rate=1.25, p=0.5),
-        PitchShift(min_semitones=-4, max_semitones=4, p=0.5),
+        TimeStretch(min_rate=0.8, max_rate=1.25, p=1.0),
+        PitchShift(min_semitones=-4, max_semitones=4, p=1.0),
     ])
 
     # Generate 2 seconds of dummy audio for the sake of example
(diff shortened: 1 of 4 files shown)


audiomentations/core/composition.py
```diff
@@ -19,59 +20,134 @@ def randomize_parameters(self, *args, **kwargs):
...
             if self.shuffle:
                 random.shuffle(transforms)
             for transform in transforms:
                 samples = transform(samples, sample_rate)
 
         return samples
 
-    def randomize_parameters(self, samples, sample_rate):
-        for transform in self.transforms:
-            transform.randomize_parameters(samples, sample_rate)
```
Cangonin, line 58: Why did you remove it ?

audiomentations/core/composition.py
```diff
@@ -19,59 +20,134 @@ def randomize_parameters(self, *args, **kwargs):
...
+
+    def randomize_parameters(self, *args, **kwargs):
+        super().randomize_parameters(*args, **kwargs)
+        self.should_apply = random.random() < self.p
+        if self.should_apply:
+            self.transform_index = random.randint(0, len(self.transforms) - 1)
+
+    def __call__(self, *args, **kwargs):
+        if not self.are_parameters_frozen:
+            self.randomize_parameters(apply_to_children=False)
```
Cangonin, line 143: So when `apply_to_children` is set to False, what does it mean here ? That you randomize parameterst to the only transform which will be applied ?

Cangonin (approved):

audiomentations/core/composition.py
iver56: It means that it only randomize the parameters (self.should_apply and self.transform_index) of "self", and not the parameters in the list self.transforms. The parameters in each transform will be randomized when it gets called.

audiomentations/core/composition.py
iver56: Because I implemented it in the base class instead :)

PR #139, opened 2021-11-08
Added SomeOf

Cangonin:
+ tests (mostly copy pasted from the tests of OneOf).

### DIFF
--- a/audiomentations/augmentations/spectrogram_transforms.py
+++ b/audiomentations/augmentations/spectrogram_transforms.py
@@ -30,7 +30,6 @@ class SpecFrequencyMask(BaseSpectrogramTransform):
         assert fill_mode in ("mean", "constant")
         self.fill_mode = fill_mode
         self.fill_constant = fill_constant
-        self.name = "SpecFrequencyMask"
 
     def randomize_parameters(self, magnitude_spectrogram):
         super().randomize_parameters(magnitude_spectrogram)
@@ -80,10 +79,6 @@ class SpecChannelShuffle(BaseSpectrogramTransform):
     """
     supports_multichannel = True
     supports_mono = False
-
-    def __init__(self):
-        super().__init__()
-        self.name = "SpecChannelShuffle"
 
     def randomize_parameters(self, magnitude_spectrogram):
         super().randomize_parameters(magnitude_spectrogram)
--- a/audiomentations/augmentations/spectrogram_transforms.py
+++ b/audiomentations/augmentations/spectrogram_transforms.py
@@ -30,6 +30,7 @@ class SpecFrequencyMask(BaseSpectrogramTransform):
         assert fill_mode in ("mean", "constant")
         self.fill_mode = fill_mode
         self.fill_constant = fill_constant
+        self.name = "SpecFrequencyMask"
 
     def randomize_parameters(self, magnitude_spectrogram):
         super().randomize_parameters(magnitude_spectrogram)
@@ -79,6 +80,10 @@ class SpecChannelShuffle(BaseSpectrogramTransform):
     """
     supports_multichannel = True
     supports_mono = False
+
+    def __init__(self):
+        super().__init__()
+        self.name = "SpecChannelShuffle"
 
     def randomize_parameters(self, magnitude_spectrogram):
         super().randomize_parameters(magnitude_spectrogram)
--- a/audiomentations/core/composition.py
+++ b/audiomentations/core/composition.py
@@ -247,14 +247,14 @@ class OneOf(BaseCompose):
 
     def __init__(self, transforms, p: float = 1.0):
         super().__init__(transforms, p)
-        self.transform_index = 0
+        self.transform_indexes = 0
         self.should_apply = True
 
     def randomize_parameters(self, *args, **kwargs):
         super().randomize_parameters(*args, **kwargs)
         self.should_apply = random.random() < self.p
         if self.should_apply:
-            self.transform_index = random.randint(0, len(self.transforms) - 1)
+            self.transform_indexes = random.randint(0, len(self.transforms) - 1)
 
     def __call__(self, *args, **kwargs):
         if not self.are_parameters_frozen:
@@ -264,7 +264,7 @@ class OneOf(BaseCompose):
         if self.should_apply:
             if "apply_to_children" in kwargs:
                 del kwargs["apply_to_children"]
-            return self.transforms[self.transform_index](*args, **kwargs)
+            return self.transforms[self.transform_indexes](*args, **kwargs)
 
         if "samples" in kwargs:
             return kwargs["samples"]
--- a/audiomentations/core/composition.py
+++ b/audiomentations/core/composition.py
@@ -1,6 +1,9 @@
 import random
 import numpy as np
-from audiomentations.core.transforms_interface import BaseWaveformTransform, BaseSpectrogramTransform
+from audiomentations.core.transforms_interface import (
+    BaseWaveformTransform,
+    BaseSpectrogramTransform,
+)
 
 
 class BaseCompose:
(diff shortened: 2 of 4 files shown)


iver56:
Ideally it should be possible to specify how many of the given transforms should run, [like in imgaug](https://github.com/aleju/imgaug/blob/0101108d4fed06bc5056c4a03e2bcb0216dac326/imgaug/augmenters/meta.py#L3188)

Examples:

```
# pick exactly two of the transforms
SomeOf(2, [transform1, transform2, transform3...])
```

```
# pick 1 to 3 of the transforms
SomeOf(1, 3 [transform1, transform2, transform3...])
```

iver56:
Would you like me to implement that, or do you want to do it?

audiomentations/core/composition.py
```diff
@@ -106,25 +107,104 @@ def __call__(self, magnitude_spectrogram):
...
+                del kwargs["apply_to_children"]
+            print(self.transform_indexes[0])
+            transformed_data = self.transforms[self.transform_indexes[0]](
+                *args, **kwargs
+            )
+
+            if hasattr(
+                self.transforms[0], "name"
+            ):  # Quick fix, added a name to the transformations on spectrograms
+                # to be able to separate transforms on the waveforms and transforms on the spectrograms
```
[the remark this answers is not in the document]
Cangonin: Yes, that's a good idea !

audiomentations/core/composition.py
```diff
@@ -106,25 +107,104 @@ def __call__(self, magnitude_spectrogram):
...
+    def __call__(self, *args, **kwargs):
+        if not self.are_parameters_frozen:
+            kwargs["apply_to_children"] = False
+            self.randomize_parameters(*args, **kwargs)
+
+        if self.should_apply:
+            if "apply_to_children" in kwargs:
+                del kwargs["apply_to_children"]
+            print(self.transform_indexes[0])
+            transformed_data = self.transforms[self.transform_indexes[0]](
```
iver56, line 160: Hmm, transformed_data can be either samples or magnitude_spectrogram.

If it is samples, that means that the next transform is also a waveform transform, which means that it expects (samples, sample_rate) as input.

Perhaps it's easier/better to write the for-loop for waveform and spectrogram independently?

And maybe check inside \_\_init\_\_ whether the given transforms are for waveform or spectrogram?

Cangonin: Yes, you're right, I haven't thought about it. :) I'll modify it as well !

iver56: Thanks :)

Cangonin:
All the comments have been adressed :)

audiomentations/core/composition.py
```diff
@@ -125,20 +143,119 @@ class OneOf(BaseCompose):
     # Augment/transform/perturb the audio data
     augmented_samples = augment(samples=samples, sample_rate=16000)
 
+    # Result: 2 or more transforms will be applied from the list of transforms.
+    ```
+
+    """
+
+    def __init__(
+        self, nbr_transforms: float or tuple, transforms, p: float = 1.0
```
iver56, line 152: I would prefer "num_transforms" over "nbr_transforms", following the naming convention used in other places in audiomentations, like

num_channels
num_frequency_bins
num_semitones
num_places_to_shift

```suggestion
        self, num_transforms: float or tuple, transforms, p: float = 1.0
```

iver56: And by the way, shouldn't this be `int or tuple` instead of `float or tuple`?

audiomentations/core/composition.py
```diff
@@ -125,20 +143,119 @@ class OneOf(BaseCompose):
...
+                    nbr_transforms_to_apply = random.randint(
+                        self.nbr_transforms[0], len(self.transforms)
+                    )
+                else:
+                    nbr_transforms_to_apply = random.randint(
+                        self.nbr_transforms[0], self.nbr_transforms[1]
+                    )
+            else:
+                nbr_transforms_to_apply = self.nbr_transforms
+            all_transforms_indexes = list(np.arange(len(self.transforms)))
```
iver56, line 174: This is no biggie, but using python's internal "range" would also work well here, avoiding the little roundtrip to numpy and back

```suggestion
            all_transforms_indexes = list(range(len(self.transforms)))
```

audiomentations/core/composition.py
```diff
@@ -125,20 +143,119 @@ class OneOf(BaseCompose):
...
+    # Augment/transform/perturb the audio data
+    augmented_samples = augment(samples=samples, sample_rate=16000)
     # Result: The audio was either time-stretched or pitch-shifted, but not both
     ```
     """
 
     def __init__(self, transforms, p: float = 1.0):
         super().__init__(transforms, p)
-        self.transform_index = 0
+        self.transform_indexes = 0
```
iver56, line 251: I feel like the plural here would imply that it's a list, but it's just an int, so I think `self.transform_index` works better here

audiomentations/core/composition.py
```diff
@@ -106,17 +111,30 @@ def __call__(self, magnitude_spectrogram):
         return magnitude_spectrogram
 
 
-class OneOf(BaseCompose):
+class SomeOf(BaseCompose):
     """
-    OneOf randomly picks one of the given transforms when called, and applies that
-    transform.
+    SomeOf randomly picks several of the given transforms when called, and applies these
```
iver56, line 116: Good docstring 👍

audiomentations/augmentations/spectrogram_transforms.py
```diff
@@ -30,6 +30,7 @@ def __init__(
         assert fill_mode in ("mean", "constant")
         self.fill_mode = fill_mode
         self.fill_constant = fill_constant
+        self.name = "SpecFrequencyMask"
```
iver56, line 33: I believe these names are not necessary now, right? See also SpecChannelShuffle

audiomentations/core/composition.py
```diff
@@ -125,20 +143,119 @@ class OneOf(BaseCompose):
...
+
+        if "samples" in kwargs:
+            return kwargs["samples"]
+        elif "magnitude_spectrogram" in kwargs:
+            return kwargs["magnitude_spectrogram"]
+        else:
+            return args[0]
+
+
+class OneOf(BaseCompose):
```
iver56, line 231: In the future, OneOf could be a subclass of SomeOf 👍

iver56 (changes requested):

audiomentations/augmentations/spectrogram_transforms.py
Cangonin: Yes, I forgot to remove them^^

audiomentations/augmentations/spectrogram_transforms.py
```diff
@@ -80,6 +81,10 @@ class SpecChannelShuffle(BaseSpectrogramTransform):
     supports_multichannel = True
     supports_mono = False
 
+    def __init__(self):
+        super().__init__()
+        self.name = "SpecChannelShuffle"
+
```
iver56, line 87: I think you have to revert this change, because I see some test failures like this now:

```
        augmenter = SpecCompose(
            [
>               SpecChannelShuffle(p=1.0),
                SpecFrequencyMask(p=1.0),
            ]
        )
E       TypeError: __init__() got an unexpected keyword argument 'p'
```

iver56:
I see 6 test failures when I run `pytest`

audiomentations/core/composition.py
Cangonin: It was to have the same name for the two of them (more consistency)  but I agree with you that the singular is more appropriate. :)

audiomentations/core/composition.py
Cangonin: Oh yeah, right !

audiomentations/core/composition.py
Cangonin: That was my initial thinking but I wasn't sure about how the randomization would work so I prefered to have this approach for a first version. I can change it now :)

audiomentations/augmentations/spectrogram_transforms.py
Cangonin: Yes, I wanted to revert it, I forgot to do it. I'm fixing it

audiomentations/core/composition.py
iver56: Yeah, it can be done in a future PR. It's not a hurry :)

Cangonin:
> I see 6 test failures when I run pytest

SomeOf works, I didn't check the others, I'm doing it now :)

Cangonin:
I don't have any failures, in which test files do you have failures ?

iver56:
All tests pass after your last commit 👍

iver56 (approved):

PR #141, opened 2021-12-15
added multichannel support for impulse response

goh-songquito:


iver56:
Thanks for the contribution :) I'll have a look

PR #142, opened 2021-12-23
Added parameter add_all_noises_with_same_level in the transform AddSh…

Cangonin:
…ortNoises.

This allows to add all the short noises at the same level. Each time that the parameters of the transform are randomised, this level takes a random value between min_snr (or rms) and max_snr (rms).

Motivation: this can simulate situations with only loud noises or only low noises and thus have a control over the loudness of the added noises.


iver56:
LGTM :) Thanks for the contribution

PR #143, opened 2022-01-13
Replace High/Low/Bandpass filters. Add BandStopFilter. Add tests, and support multichannel filters.

mmxgn:
What the title says. This PR:

- Removes `pydub` as a requirement for filter transforms.
- Adds a `ButterworthFilter` class where all filters inherit from to reduce code redudancy.
- Adds magnitude response tests to make sure that the transforms do what we think they do.
- Speeds up by almost 60X the application of `LowPassFilter`, `HighPassFilter`, `BandPassFilter`, compared to before.
- Applies the transform to all channels in the case of a multichannel audio while taking in consideration initial conditions instead of just assuming channels start from 0.
- Allows for zero-phase filtering to avoid smearing transients (but will result in a 6dB drop at cutoff frequency instead of 3dB in the non zero-phase case).
- Adds a `BandStopFilter` which can be used for e.g. frequency masking.

For the performance claims, this is with master:
image

And this is with this PR:
image


README.md
```diff
@@ -380,6 +380,12 @@ _The following table is valid for new versions of audiomentations, like >=0.18.0
 
 ## Unreleased
 
+### Added
+* Added magnitude response tests for `BandPassFilter`, `LowPassFilter`, `HighPassFilter`
+* Added `BandStopFilter`
+### Changed
+* Changed `BandPassFilter`, `LowPassFilter`, `HighPassFilter`, `BandStopFilter` to use scipy's butterworth filters. Now they support multichannel as well.
```
iver56, line 387: You could also mention that it's much faster :)

iver56:
Nice!

Quick question: How can we control the steepness? 6 dB per octave doesn't cut it, so I want to increase the default to 12 or 24 dB per octave

mmxgn:
> Quick question: How can we control the steepness? 6 dB per octave doesn't cut it, so I want to increase the default to 12 or 24 dB per octave

Ah, I put 6db (order of 1) since that was what it was before. Increasing the order to 2 increases it to 12dB/octave, etc.. but also affects the phase more. I will look it up to change tomorrow together with the rest.

audiomentations/augmentations/transforms.py
```diff
@@ -1692,6 +1724,7 @@ def __init__(
             Must be a multiple of 6
         :param max_rolloff: Maximum filter roll-off (in db/octave)
             Must be a multiple of 6
+        :param zero_phase: Whether filtering should be zero phase
```
iver56, line 1727: It would be nice to have a sentence of two that explain the implications of this in a way most developers can understand.

E.g. will audio quality improve (audibly or just in theory)? Will it be slower? By how much?

iver56:
Nice! I'll give this a listen

iver56:
Thanks for the patience 😅 I wasn't able to complete the review yesterday. I'll see what I can do today.

iver56:
The filters sound good :) However I believe that the BPF and BSF default params could use a little tweaking.

* BPF max center freq could be higher
* It sounds like BSF in average is not doing enough to the input

But those changes could be a separate PR

When comparing exec time numbers in demo.py I find that the new augmentations are at least 22x faster. Nice! :)

iver56 (approved):

Issue #144, opened 2022-01-20
Tweak BPF and BSF defaults and express bandwidth as a ratio of the center freq


iver56:
- [x] Adopt this approach: https://github.com/asteroid-team/torch-audiomentations/blob/master/torch_audiomentations/augmentations/band_pass_filter.py#L18

iver56:
Also test that passing `p` works as expected

iver56:
Also tweak PeakingFilter defaults - make it more aggressive

iver56:
And figure out `FrequencyMask` vs. `BandStopFilter` - they do roughly the same thing now. Maybe deprecate FrequencyMask? Or make it a super class of BandStopFilter?

iver56:
Make sure that the upper frequency doesn't exceed the nyquist frequency, or else scipy will get cross

iver56:
use mel scale in PeakingFilter

iver56:
- [ ] Pick cutoff frequency in mel scale for the shelving filters
- [ ] Also, is "center frequency" the correct terminology for the shelving filters?

PR #147, opened 2022-02-07
Tweak filters (prepare for release)

iver56:
Use bandwidth fraction (relative to center frequency) instead of bandwidth directly.

I also tweaked the default values. And I added a few guards. And I tweaked the way frequencies are selected (mel instead of linear). And I fixed some docstrings and the default parameters of the LPF.


iver56:
Close #144

Issue #123, opened 2021-07-22
Padding transform


iver56:
This transform takes a fraction of the end or the start of the audio and treats that part as padding. We can implement several modes:

* constant (zero)
* edge - pads with the edge values of array
* wrap
* reflect

iver56:
See the explanation of modes here:

https://numpy.org/doc/stable/reference/generated/numpy.pad.html

I think wrap can also in some cases be called circular.

Hope this helps

PR #159, opened 2022-02-21
Add Padding

atamazian:
Attempt to implement https://github.com/iver56/audiomentations/issues/123.

### DIFF
--- a/audiomentations/augmentations/padding.py
+++ b/audiomentations/augmentations/padding.py
@@ -48,10 +48,16 @@ class Padding(BaseWaveformTransform):
             samples = samples[..., -skip_idx:]
        
         pad_width = orig_len - samples.shape[-1]
+        
         if n_channels > 1:
             if r < 0.5:
                 pad_width =  ((0, 0)*(n_channels-1), (pad_width, 0))
             else:
                 pad_width =  ((0, 0)*(n_channels-1), (0, pad_width))
+        else:
+            if r < 0.5:
+                pad_width = (pad_width, 0)
+            else:
+                pad_width = (0, pad_width)
         samples = np.pad(samples, pad_width, self.mode)
         return samples
--- a/audiomentations/augmentations/padding.py
+++ b/audiomentations/augmentations/padding.py
@@ -11,6 +11,10 @@ class Padding(BaseWaveformTransform):
     supports_multichannel = True
     
     def __init__(self, mode='constant', p=0.5):
+        """
+        :param mode: Padding mode. Must be one of 'constant', 'edge', 'wrap', 'reflect'
+        :param p: The probability of applying this transform
+        """
         super().__init__(p)
         self.mode = mode
 
--- a/audiomentations/augmentations/padding.py
+++ b/audiomentations/augmentations/padding.py
@@ -39,7 +39,7 @@ class Padding(BaseWaveformTransform):
         a = self.min_fraction*orig_len
         b = self.max_fraction*orig_len
         
-        skip_idx = int(np.random.uniform(a, b))
+        skip_idx = int(np.ceil(np.random.uniform(a, b)))
         r = np.random.random()
         if r < 0.5:
             samples = samples[..., :skip_idx]
--- a/audiomentations/augmentations/padding.py
+++ b/audiomentations/augmentations/padding.py
@@ -40,19 +40,18 @@ class Padding(BaseWaveformTransform):
         b = self.max_fraction*orig_len
         
         skip_idx = int(np.ceil(np.random.uniform(a, b)))
+   
         r = np.random.random()
         if r < 0.5:
             samples = samples[..., :skip_idx]
         else:
             samples = samples[..., -skip_idx:]
-            
+       
+        pad_width = orig_len - samples.shape[-1]
         if n_channels > 1:
-            pad_width =  ((0, 0)*(n_channels-1), (orig_len, orig_len - samples.shape[-1]))
-        else:
-            pad_width = orig_len - samples.shape[-1]
+            if r < 0.5:
+                pad_width =  ((0, 0)*(n_channels-1), (pad_width, 0))
+            else:
+                pad_width =  ((0, 0)*(n_channels-1), (0, pad_width))
         samples = np.pad(samples, pad_width, self.mode)
-        if r < 0.5:
-            samples = samples[..., :orig_len]
-        else:
-            samples = samples[..., -orig_len:]
         return samples
--- a/audiomentations/augmentations/padding.py
+++ b/audiomentations/augmentations/padding.py
@@ -19,13 +19,13 @@ class Padding(BaseWaveformTransform):
         skip_idx = np.random.randint(1, orig_len-1)
         r = np.random.random()
         if r < 0.5:
-            samples = samples[:skip_idx]
+            samples = samples[..., :skip_idx]
         else:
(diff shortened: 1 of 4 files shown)


iver56:
Hi :) Thanks for picking up this task! I have some feedback on your initial attempt:

The original idea was to pick a part of the start of the end (with random length), then discard that, and then pad the remaining audio so that it exactly compensate for the discarded part. In the end, the output audio (augmented audio) has the same length as the input audio.

Here's an example with illustration:

Pick an end time (pink vertical line). Then pretend that the pink part of the audio is the audio to pad (i.e. "discard" the gold-colored part). And then replace the gold-colored part of the audio with padding.

bilde

I can try to explain the main motivation for having such a transform. Some audio deep learning model architectures expect a fixed input length, for example 10 seconds. Then what happens if a user has a shorter audio? That shorter audio would have to be extended to 10 seconds to make it work in the model. There are several ways of doing this (fill in a zeroes, wrap, reflect, repeat...). If the model is trained on this particular way of getting input audio, then it will likely perform better on those short sounds.

If you could align your implementation with this idea, that would be nice!

audiomentations/augmentations/padding.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+    
+    supports_multichannel = True
+    
+    def __init__(self, mode='constant', p=0.5):
+        super().__init__(p)
+        self.mode = mode
+
+    def apply(self, samples, sample_rate):
+        orig_len = len(samples)
```
iver56:
```suggestion
        orig_len = samples.shape[-1]
```

for multichannel support

audiomentations/augmentations/padding.py
```diff
@@ -0,0 +1,31 @@
...
+        skip_idx = np.random.randint(1, orig_len-1)
+        r = np.random.random()
+        if r < 0.5:
+            samples = samples[:skip_idx]
```
iver56, line 22:
```suggestion
            samples = samples[..., :skip_idx]
```

for multichannel support

iver56 (changes requested):
It could use a test that checks the multichannel (e.g. stereo) case works as expected

audiomentations/augmentations/padding.py
atamazian: Already done)

audiomentations/augmentations/padding.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def apply(self, samples, sample_rate):
+        orig_len = samples.shape[-1]
+        skip_idx = np.random.randint(1, orig_len-1)
+        r = np.random.random()
+        if r < 0.5:
+            samples = samples[..., :skip_idx]
+        else:
+            samples = samples[..., -skip_idx:]
+        pad_width = orig_len - len(samples)
```
iver56, line 25:
```suggestion
        pad_width = orig_len - samples.shape[-1]
```

audiomentations/augmentations/padding.py
atamazian: Also done

audiomentations/augmentations/padding.py
atamazian: Also done

audiomentations/augmentations/padding.py
```diff
@@ -0,0 +1,31 @@
...
+class Padding(BaseWaveformTransform):
+    """
+    Apply padding to the audio signal -  take a fraction of the end or the start of the audio
+    and treat that part as padding.
+    """
+    
+    supports_multichannel = True
+    
+    def __init__(self, mode='constant', p=0.5):
+        super().__init__(p)
```
iver56, line 14: Write in a docstring what values can be used for `mode`

atamazian: How about allowing used of all numpy.pad modes, not just 4 mentioned earlier?

atamazian:
How about allowing usege of all numpy.pad modes, not just 4 mentioned earlier? I can write "Must be one of numpy.pad allowed modes" instead of "Must be one of 'constant', 'edge', 'wrap', 'reflect'"

iver56:
> How about allowing usage of all numpy.pad modes, not just 4 mentioned earlier? I can write "Must be one of numpy.pad allowed modes" instead of "Must be one of 'constant', 'edge', 'wrap', 'reflect'"

I can appreciate the idea, but I think that would add an extra burden to the user (the need to look up the numpy documentation), and also not all pad modes make sense in the audio/waveform space

atamazian:
I see. How about adding assert condition then? To prevent user from selecting unsuitable mode.

atamazian:
I think I know a reason for that. numpy.pad pads multi-dimensional array by all axes, while we need padding only on axis X.

atamazian:
Try now.

iver56:
Yep, that did the trick :)

Btw, I made a PR for you - it updates the demo script in your branch

https://github.com/atamazian/audiomentations/pull/1/files

I ran the demo script, and I think I identified a weakness in the transform: 3 out of 5 of the stereo files (when running with mode="constant") ended up becoming completely silent. Ideally, it should silence a _part of the audio_, but not the whole audio

bilde

atamazian:
Got it, I'll check it out. By the way, which limits you think would be the best for replacing signal's part?

atamazian:
I've added `min_fraction` and `max_fraction` parameters.

iver56:
Nice, I think that's better, but it still has issues where it takes away too much. E.g. see here, output from the demo script:

bilde

Maybe you can run the demo script on your end as well to reproduce it?

atamazian:
I'm working on it.

atamazian:
I'm also getting zero signals in some runs. I think it's because in some cases entire signal ends up skipped before padding.

atamazian:
Can you try now? I've rewritten the code, should be no such cases now.

iver56:
I agree, but now there seems to be a different issue: the outputs have different length

bilde

iver56:
I would definitely prefer having a output length, since this transform is mainly useful for models with constant input length. In other words, they can't use inputs of varying length. If you want, I can merge this now and massage it a bit some time before it gets released. Would you be okay with that?

atamazian:
I think the ouput array has the same length, as input, just part of it from left or right has zero values. Can you try `edge` option? It seems that you use some visualization tool I don't have.

atamazian:
As for merging, let's ensure that this transform works absolutely as intended. There's no need to rush.

iver56:
It's a program called Audacity :) It's free. You can drag in audio files there to visualize and play them.

iver56:
I tried with `edge` now, and it has the same issue

atamazian:
Great, I'll install it. I knew about it's existence, but never used.

atamazian:
OK, in that case it's an error. I'll look into it.

iver56:
It would be nice with a shape similarity check in the tests. I.e. check that the shape of the input ndarray is the same as the output ndarray

atamazian:
I found the error - pad_width was incorrectly calculated for mono audio. I've fixed that. As for test shape similarity check - it was present. It's just a forgot to check mono( I'll be more careful next time.

iver56:
Thanks for the contribution. I'll merge this now. As I said earlier, I might massage it a little bit before it gets released, so I'm confident that it's ready for prime time 😄

iver56 (approved):

Issue #150, opened 2022-02-10
Add unit tests for LowPassFilter


iver56:

iver56:
@mmxgn Do you want to do this?

PR #162, opened 2022-02-28
Add unit test for LowPassFilter transform

mmxgn:
What the title says. Copied and pasted from  TestHighPassFilterTransform but changed frequencies of interest, gains of interest, and tolerances accordingly.


iver56:
Nice :) This would close #150

PR #167, opened 2022-03-01
Add multichannel audio tests

mmxgn:
- Extend the tests to cover n channel cases instead of just two.
- Add missing multichannel test for the Lowpass filter


iver56:
Thanks :) This would be easier to get merged if it didn't also include the RoomSimulator code

mmxgn:
> Thanks :) This would be easier to get merged if it didn't also include the RoomSimulator code

Oops.

iver56 (approved):
Nice :)

PR #170, opened 2022-03-04
Add parameter signal_gain_in_db_during_noise in AddShortNoises

### LINKED ISSUE #160: Add "signal_gain_during_noise" in AddShortNoises
If we do this, AddShortNoises can be used to implement [InterruptPulse](https://github.com/iver56/audiomentations/pull/122) (by setting `signal_gain_during_noise` to some very low value)

It can then also be used to replace parts of the original signal with other sounds that should be considered signal

iver56:
This new feature can temporarily turn down the gain of the signal during the addition of a short noise. If we set the new gain param to a very low level, it can be used to _replace_ the signal with the noise. That could happen e.g. if a microphone wire gets disconnected (e.g. the singing vocal gets replaced by a buzzing sound)

Close #160


audiomentations/augmentations/add_short_noises.py
```diff
@@ -305,30 +325,35 @@ def apply(self, samples, sample_rate):
...
                     noise_samples = noise_samples * gain
 
-                    noise_placeholder[
-                        start_sample_index:end_sample_index
-                    ] += noise_samples
-        # Return a mix of the input sound and the added sounds
-        return samples + noise_placeholder
+                noise_placeholder[start_sample_index:end_sample_index] += noise_samples
+                if gain_signal:
+                    signal_mask[start_sample_index:end_sample_index] = np.maximum(
```
Cangonin, line 344: noise_gain only takes positive values from my understanding. And signal_mask contains only 0.  The np.maximum is not so useful here, no ?

tests/utils.py
```diff
@@ -19,3 +19,20 @@ def plot_matrix(matrix, output_image_path=None, vmin=None, vmax=None, title=None
     else:
         plt.show()
     plt.close(fig)
+
+
+def plot_waveforms(wf1, wf2=None, wf3=None, title="Untitled"):
```
Cangonin, line 24: I agree that this is useful. Where do you use it ?

Cangonin (approved):
Looks good to me. :)

audiomentations/augmentations/add_short_noises.py
iver56: Hmm, I think this is correct. It has to do with how we deal with the case when noises overlap. Let me try to illustrate

bilde

I'll do a few runs with plots and overlap to be sure

tests/utils.py
iver56: I used it temporarily in my code to check some things before I removed it again ^^ I might want to use it for development and debugging later

audiomentations/augmentations/add_short_noises.py
Cangonin: Oh yes, I see, I haven't thought about that case !

PR #178, opened 2022-03-14
Implement equalizer

### LINKED ISSUE #12: Add equalizer transformation

iver56:
Close #12


mmxgn:
Nice I will be reviewing it today.

audiomentations/augmentations/equalizer.py
```diff
@@ -0,0 +1,94 @@
...
+    supports_multichannel = True
+
+    def __init__(
+        self,
+        min_gain_in_db: float = -12.0,
+        max_gain_in_db: float = 12.0,
+        p: float = 0.5,
+    ):
+        """
+        :param p: The probability of applying this transform
```
mmxgn, line 21: `min_gain_in_db`, `max_gain_in_db` missing from the docstring

audiomentations/augmentations/equalizer.py
```diff
@@ -0,0 +1,94 @@
+from audiomentations import LowShelfFilter, PeakingFilter, HighShelfFilter
+from audiomentations.core.transforms_interface import BaseWaveformTransform
+
+
+class Equalizer(BaseWaveformTransform):
+    """
+    Adjust the volume of different frequency bands. This transform is a 7-band
```
mmxgn, line 7: Why is it 7 band? Where is it based from? How is this going to be used?

Equalizers there have different needs for different reasons. For example this paper: 

"Neural Parametric Equalizer Matching Using Differentiable Biquads" by Nercessian they use a 12 band (10 peaking + LS + HS).

In "Neural Third Octave Graphic Equalizer" by Rämö and Välimäki they use 33 bands. 

On the other hand in the IIRNet paper (DIRECT DESIGN OF BIQUAD FILTER CASCADES WITH DEEP LEARNING BY SAMPLING RANDOM POLYNOMIALS by Colonel et al) they use a variable order. 

I would like it to be of parametrized order. 

A suggestion:

Parameters as:

Two booleans for whether we have Lowshelf/Highshelf filters.
An order parameter which controls the number of peaking filters. 
Arrays for center frequencies gains, and qs.

audiomentations/augmentations/equalizer.py
```diff
@@ -0,0 +1,94 @@
...
+        max_gain_in_db: float = 12.0,
+        p: float = 0.5,
+    ):
+        """
+        :param p: The probability of applying this transform
+        """
+        super().__init__(p)
+        assert min_gain_in_db <= max_gain_in_db
+
+        self.low_shelf_filter = LowShelfFilter(
```
mmxgn, line 26: How were those options chosen?

audiomentations/augmentations/equalizer.py
```diff
@@ -0,0 +1,94 @@
...
+            ),
+            PeakingFilter(
+                min_center_freq=2936.0 / 1.5,
+                max_center_freq=2936.0 * 1.5,
+                min_gain_db=min_gain_in_db,
+                max_gain_db=max_gain_in_db,
+                p=1.0,
+            ),
+        ]
+        self.high_shelf_filter = HighShelfFilter(
```
mmxgn, line 70: What about higher frequencies?

mmxgn (changes requested):

audiomentations/augmentations/equalizer.py
iver56: > Why is it 7 band? Where is it based from?

It is inspired by Fruity Parametric EQ 2, a plugin that I have personally used a lot, and I like it a lot. It covers all my EQ needs. https://github.com/iver56/audiomentations/issues/12#issuecomment-881243622

> How is this going to be used?

Because it changes the timbre, but keeps the overall "class" of the sound the same, it can be used for data augmentation to make ML models more robust to various frequency spectrums. Many things can affect the spectrum, like room acoustics, any objects that are between the microphone and the sound source, microphone type/model, distance between sound source and microphone...

> Equalizers there have different needs for different reasons. For example this paper:
> 
> "Neural Parametric Equalizer Matching Using Differentiable Biquads" by Nercessian they use a 12 band (10 peaking + LS + HS).
> 
> In "Neural Third Octave Graphic Equalizer" by Rämö and Välimäki they use 33 bands.
> 
> On the other hand in the IIRNet paper (DIRECT DESIGN OF BIQUAD FILTER CASCADES WITH DEEP LEARNING BY SAMPLING RANDOM POLYNOMIALS by Colonel et al) they use a variable order.

Yes, I realize that I cannot/won't cover everyone's EQ needs with this class. It's meant to be a "good enough" general purpose thing.

> parametrized order

Do you mean the number of filters in the chain? Or the order of application?

audiomentations/augmentations/equalizer.py
iver56: I used the same center freq means as the defaults in Fruity Parametric EQ 2. The min-max values were used to give them some wiggle room (a "perfect fifth" up or down) without having too much overlap potential between the filters

audiomentations/augmentations/equalizer.py
iver56: The high-shelf filter will affect all frequencies above the chosen center freq, so essentially all higher frequencies get gained too. Also, I didn't want to do too much above 8 kHz because a lot of folks use 16 kHz audio.

iver56:
I've included the new transform in the demo script, in case you want to listen to the outputs :)

audiomentations/augmentations/equalizer.py
mmxgn: >Yes, I realize that I cannot/won't cover everyone's EQ needs with this class. It's meant to be a "good enough" general purpose thing.

In that case I would be more comfortable with a name that suggests what type of EQ it is.

E.g. `SevenBandGraphicEQ`.

Additionally would be helpful the list of frequencies and Qs in the class docstring since those do not change.

> Do you mean the number of filters in the chain? Or the order of application?

I meant the bands. For example if K=12 the Equalizer can have 

1 LowShelf, + 10 Peaking Filters + 1 Highshelf.

And could center the filters accordingly (so they do not overlap and extend to full band).

A more generic EQ could be better implemented as a transform directly.

iver56: Ok, I hear you. I think I'll do this:

* Rename it to `SevenBandParametricEQ` (because frequencies and Qs are chosen at random, it's more a parametric EQ than a graphic EQ)
* Add parameters `use_low_shelf_filter` and `use_high_shelf_filter` (bool)
* List min/max center freq of each band in docstring
* Make an issue for a new class `ParametricEQ` with tunable K

iver56:
- [ ] Rename `min_gain_in_db` to `min_gain_db` (and similar for max)

audiomentations/augmentations/equalizer.py
mmxgn: Sounds like a plan.

PR #189, opened 2022-05-12
Add an Air Absorption "Simulator"

mmxgn:
What it says on the Tin:

A quick and dirty way to simulate air absorption (NOT DISTANCE or REFLECTIONS). 

Basically uses the STFT to emulate a low pass filter with variable octave attenuation and the attenuation coefficients from pyroomacoustics.


audiomentations/augmentations/wind_absorption.py
```diff
@@ -0,0 +1,167 @@
...
+                key += f"_{bounds[n-1]}-{bounds[n]}%"
+                break
+
+        # Convert to attenuations
+        attenuation_values = np.exp(
+            -distance * np.array(self.air_absorption_table[key])
+        )
+
+        # Frequencies to calculate the attenuations caused by wind absorption
+        frequencies = librosa.fft_frequencies(sr=48000, n_fft=256)
```
mmxgn, line 128:
```suggestion
        frequencies = librosa.fft_frequencies(sr=48000, n_fft=self.n_fft)
```

audiomentations/augmentations/wind_absorption.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def __init__(
+        self,
+        min_temperature=10.0,
+        max_temperature=10.0,
+        min_humidity=30,
+        max_humidity=90,
+        min_distance=10,
+        max_distance=100,
+        n_fft=256,
```
iver56, line 59: There's a big chance the user won't take the time to understand this and set it correctly. Maybe it can be decided automatically (by a specific fraction of the sample rate, then ceil to nearest power of two) instead of being a parameter?

audiomentations/augmentations/wind_absorption.py
```diff
@@ -0,0 +1,167 @@
+from audiomentations.core.transforms_interface import BaseWaveformTransform
+import numpy as np
+import librosa
+
+
+class WindAbsorption(BaseWaveformTransform):
```
iver56, line 6: Rename to `AirAbsorption`?

audiomentations/augmentations/wind_absorption.py
```diff
@@ -0,0 +1,167 @@
...
+    """
+    Applies a wind absorption transform parametrized by temperature, humidity, and distance
```
iver56, line 8: Mention that it resembles a low pass filter

audiomentations/augmentations/wind_absorption.py
```diff
@@ -0,0 +1,167 @@
...
+    This is not a scientifically accurate transform but basically applies a uniform
+    filterbank with attenuations given by:
+
+    att = exp(- distance * absorption_coefficient)
+
+    where distance is the microphone-source assumed distance in meters and `absorption_coefficient`
+    is adapted from a look up table by pyroomacoustics [1].
+
+    Note: This only "simulates" room absorption, and does not apply the distance law
+    (use a GainTransform for that)
```
iver56, line 20: Note: This only "simulates" the dampening of high frequencies, and does not gain down the audio according to the distance law.
Gain augmentation needs to be done separately

audiomentations/augmentations/wind_absorption.py
```diff
@@ -0,0 +1,167 @@
...
+    Applies a wind absorption transform parametrized by temperature, humidity, and distance
+    of source.
+
+    This is not a scientifically accurate transform but basically applies a uniform
+    filterbank with attenuations given by:
+
+    att = exp(- distance * absorption_coefficient)
+
+    where distance is the microphone-source assumed distance in meters and `absorption_coefficient`
+    is adapted from a look up table by pyroomacoustics [1].
```
iver56, line 17:
```suggestion
    is adapted from a lookup table by pyroomacoustics [1].
```

audiomentations/augmentations/wind_absorption.py
```diff
@@ -0,0 +1,167 @@
...
+        "20C_70-90%": [
+            x * 1e-3 for x in [0.1, 0.3, 0.6, 1.1, 1.7, 3.5, 10.6, 31.2, 93.8]
+        ],
+        "center_freqs": [125, 250, 500, 1000, 2000, 4000, 8000, 16000, 32000],
+    }
+
+    def __init__(
+        self,
+        min_temperature=10.0,
+        max_temperature=10.0,
```
iver56, line 54:
```suggestion
        max_temperature=20.0,
```

audiomentations/augmentations/wind_absorption.py
```diff
@@ -0,0 +1,167 @@
...
+        min_humidity=30,
```
iver56, line 55: Some type hints would be nice

audiomentations/augmentations/wind_absorption.py
iver56, line 128: sr needs to be set to sample_rate

audiomentations/augmentations/wind_absorption.py
```diff
@@ -0,0 +1,167 @@
...
+
+        # Apply using STFT
+        if len(samples.shape) == 1:
+            stft = librosa.stft(samples, n_fft=self.n_fft)
+
+            # Compute mask
+            mask = np.tile(linear_target_attenuations, (stft.shape[1], 1)).T
+
+            # Compute target degraded audio
+            degraded_audio = librosa.istft(stft * mask)
```
iver56, line 149: check if it's possible to specify a desired length to the istft function instead of doing the min_length trick

tests/test_wind_absorption.py
```diff
@@ -0,0 +1,38 @@
...
+
+class TestWindAbsorptionTransform:
+    @pytest.mark.parametrize("temperature", [10, 20])
+    @pytest.mark.parametrize("humidity", [30, 50, 70, 90])
+    @pytest.mark.parametrize("distance", [5, 10, 20, 100])
+    @pytest.mark.parametrize("sample_rate", [8000, 16000, 48000])
+    def test_one_single_input(self, temperature, humidity, distance, sample_rate):
+        np.random.seed(1)
+
+        samples = get_chirp_test(sample_rate, 10)
```
iver56, line 25: Test both 1D and 2D inputs

iver56 (changes requested):

audiomentations/augmentations/air_absorption.py
```diff
@@ -0,0 +1,172 @@
...
+    """
+    taken jhoyla's answer here:
+    https://stackoverflow.com/questions/14267555/find-the-smallest-power-of-2-greater-than-or-equal-to-n-in-python
+    """
+    return 1 if x == 0 else 2 ** (x - 1).bit_length()
+
+
+class AirAbsorption(BaseWaveformTransform):
+    """
+    Applies a wind absorption transform parametrized by temperature, humidity, and distance
```
Cangonin, line 16: Is it a wind or an air absorption transform ?

audiomentations/augmentations/air_absorption.py
```diff
@@ -0,0 +1,172 @@
...
+
+    [1] https://github.com/LCAV/pyroomacoustics
+    """
+
+    supports_multichannel = True
+
+    # Table of air absorption coefficients adapted from `pyroomacoustics`.
+    # The last two columns have been extrapolated with `curve_fit`
+    air_absorption_table = {
+        "10C_30-50%": [
```
Cangonin, line 39: Maybe specify what the different parts of the name "10C_30-50%" mean ? I guess that it is temperature, distance and humidity percentage, bit I think it would be more clear if you specify that as a comment.

audiomentations/augmentations/air_absorption.py
```diff
@@ -0,0 +1,172 @@
...
+        p=0.5,
+    ):
+        """
+        :param min_temperature: Minimum temperature in celcius
+        :param max_temperature: Maximum temperature in celsius
+        :param min_humidity: Minimum humidity in percent %
+        :param max_humidity: Maximum humidity in percent %
+        :param min_distance: Minimum microphone-source distance in meters.
+        :param max_distance: Maximum microphone-source distance in meters.
+        :param n_fft: The number of bins in the estimated filterbank
```
Cangonin: nfft is not present anymore in the parameters, you can remove it from the docstring.

audiomentations/augmentations/air_absorption.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def apply(self, samples, sample_rate):
+        assert samples.dtype == np.float32
+
+        humidity = self.parameters["humidity"]
+        distance = self.parameters["distance"]
+
+        # Choose correct absorption coefficients
+        key = str(int(self.parameters["temperature"])) + "C"
+        bounds = [30, 50, 70, 90]
```
Cangonin, line 121: So technically, we can choose only some values for the humidity and distance according to what is present in the table. I think it should be specified that we use only coefficients in `air_absorption_table`  in the docstring of this transform so that the user knows about it.

tests/test_air_absorption.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def get_chirp_test(sample_rate, duration):
+    """Create a `duration` seconds chirp from 0Hz to `nyquist frequency`"""
+    n = np.arange(0, duration, 1 / sample_rate)
+    samples = scipy.signal.chirp(n, 0, duration, sample_rate // 2, method="linear")
+    return samples.astype(np.float32)
+
+
+class TestWindAbsorptionTransform:
```
Cangonin, line 17: You could maybe also test that you chose the right key in `air_absorption_table` ?

audiomentations/augmentations/air_absorption.py
mmxgn: Already did.

audiomentations/augmentations/air_absorption.py
mmxgn: This is something the user shouldn't see. I specify what the values are in the docstring.

audiomentations/augmentations/air_absorption.py
mmxgn: Air, sorry.

tests/test_air_absorption.py
mmxgn: Can you give an example of what you mean?

audiomentations/augmentations/air_absorption.py
mmxgn: The user can choose any value between 30 and 90 and the transform translates it to the relevant key. I will add a bit more about the boundaries though.

Cangonin: I mean that the user should know that the transform translates it to the relevant key. :) If I were to use this transform, I would like to know if a transform maps my input to something else.

mmxgn: I already specify I am using a look up table by pyroomacoustics.

tests/test_air_absorption.py
Cangonin: I mean that you map the input to the relevant key

audiomentations/augmentations/air_absorption.py
```diff
@@ -0,0 +1,171 @@
...
+        :param min_humidity: Minimum humidity in percent %
+        :param max_humidity: Maximum humidity in percent %
+        :param min_distance: Minimum microphone-source distance in meters.
+        :param max_distance: Maximum microphone-source distance in meters.
+        :param p: The probability of applying this transform
+        """
+        assert float(min_temperature) in [
+            10.0,
+            20.0,
+        ], "Sorry, the only supported temperatures are either 10 or 20 degrees celcius"
```
Cangonin, line 82: Why do you assert that the temperature can only take two values ? You could do as you already do for the humidity and the distance, ie directly map the temperature value it to the keys in the table. For example, if the user choose 12, you translate it to 10.

tests/test_air_absorption.py
mmxgn: The key is not exposed though and I would prefer not to.

audiomentations/augmentations/air_absorption.py
mmxgn: > You could do as you already do for the humidity and the distance, ie directly map the temperature value it to the keys in the table. 

If a better model is added then that would be the case but here the keys have been chosen for distinct temperatures but ranges of humidity so I keep it like that to be used as originally intended.

audiomentations/augmentations/wind_absorption.py
mmxgn: I changed it so that its computed automatically so that the first octave band (0-250Hz) can be shown as a single fft bin.

audiomentations/augmentations/wind_absorption.py
mmxgn: I used `np.ndarray` for arrays because I don't want to force a newer version of numpy just for the type hinting.

audiomentations/augmentations/air_absorption.py
Cangonin: I still think it would be more clear if you would say something like:
# Table of air absorption coefficients adapted from `pyroomacoustics`.
# The keys describe the temperature and the humidity range.
# The last two columns have been extrapolated with `curve_fit`

At first sight, I didn't understand that 30-50% was a range, I thought it was two different values.

mmxgn: I see what you mean. Will fix it.

tests/test_air_absorption.py
iver56, line 17:
```suggestion
class TestAirAbsorptionTransform:
```

audiomentations/augmentations/air_absorption.py
```diff
@@ -0,0 +1,177 @@
...
+        attenuation_values = np.exp(
+            -distance * np.array(self.air_absorption_table[key])
+        )
+
+        # Calculate n_fft so that the lowest band can be stored in a single
+        # fft bin.
+        first_band_bw = self.air_absorption_table["center_freqs"][0] * 2
+        n_fft = next_power_of_2(int(sample_rate / 2 / first_band_bw))
+
+        # Frequencies to calculate the attenuations caused by wind absorption
```
iver56, line 143:
```suggestion
        # Frequencies to calculate the attenuations caused by air absorption
```

iver56 (approved):

audiomentations/augmentations/air_absorption.py
```diff
@@ -0,0 +1,177 @@
...
+                break
+
+        # Convert to attenuations
+        attenuation_values = np.exp(
+            -distance * np.array(self.air_absorption_table[key])
+        )
+
+        # Calculate n_fft so that the lowest band can be stored in a single
+        # fft bin.
+        first_band_bw = self.air_absorption_table["center_freqs"][0] * 2
```
Cangonin, line 140: If your first center frequency is 125 Hz, then your octave band bandwidth is 88 - 176 Hz, which is not 125*2. Or maybe `first_band_bw `refers to something else ?

mmxgn: You are right. Nice catch.

Cangonin (approved):

PR #194, opened 2022-05-20
Moved the calculation of noise_rms in the background noise transform

Cangonin:
Fixed a bug where the rms of the background nosie would be calculated before applying oprtional transforms to the background noise.

Should I add a test for that ?


iver56:
Thanks for the fix :) A test would indeed be nice, if you have time

iver56 (approved):

Cangonin:
I will add a test later, I guess it will take some time to write. :)

Issue #186, opened 2022-05-11
Pick sample rate logarithmically instead of linearly in Resample


iver56:
And maybe rename it to ReSpeed or something? That would make its purpose more clear

iver56:
ReSpeed (or maybe just Speed) should have parameters like min_speed=0.8, max_speed=1.25 and then it picks the sample rate based on that.

And then what happens with Resample? I don't know. Do we just deprecate it and eventually remove it? Or do we remove the min/max parameters and let the user specify a single target sample rate?

PR #195, opened 2022-05-21
Add sample rate augmentation

WeberJulian:
TimeStretch and Pitch augment can create heavy artifacts.
For some use cases it might be better to change the speed by changing the sample-rate even if that entails creating wrong correlation between pitch and speed.

If you agree with the usefulness, I'll add tests and I'm also opened to name change as I was not very inspired haha


WeberJulian:
nevermind, I thought resample augmentation was to simulate the artefacts of resampling but that the audio was resampled at the original samplerate.

iver56:
I agree with the usefulness. I have some ideas for how the current Resample transform can be improved:
https://github.com/iver56/audiomentations/issues/186

One of the things I want is, as you have suggested, `min_rate=0.8, max_rate=1.25`

Issue #181, opened 2022-03-17
Implement GainPeak


iver56:
Something that turns the gain up and then down, almost like a sound source that drives by

This is a specialization of https://github.com/iver56/audiomentations/issues/53#issuecomment-1038862832

atamazian:
Which parameters should be available to be set? Max gain and its position?

iver56:
I guess it should randomize at least these two things:

* the difference between min gain and max gain
* "the wideness" of the gain curve (think like: how fast does the sound source drive by)

Bonus:

* The shape of the gain curve. Could be triangle, could be gaussian, could be cosine...

atamazian:
Can you assign this issue to me? I'll try to do it.

iver56:
Great! Done

atamazian:
Should we let user specify `min_gain`, or should we set it to some constant value?

iver56:
I guess the most important thing is the difference between the minimum gain and the maximum gain.

In the future I want to establish a common post-gain pattern that can compensate: https://github.com/iver56/audiomentations/issues/171

Thanks for the PR by the way! I will have a look at soon-ish :)

atamazian:
Thanks! As for now, I've `min_gain` specified by the user.

PR #200, opened 2022-06-06
Add gain peak

atamazian:
Attempt to implement https://github.com/iver56/audiomentations/issues/181.

### DIFF
--- a/audiomentations/__init__.py
+++ b/audiomentations/__init__.py
@@ -9,6 +9,7 @@ from .augmentations.clip import Clip
 from .augmentations.clipping_distortion import ClippingDistortion
 from .augmentations.frequency_mask import FrequencyMask
 from .augmentations.gain import Gain
+from .augmentations.gain_peak import GainPeak
 from .augmentations.gain_transition import GainTransition
 from .augmentations.high_pass_filter import HighPassFilter
 from .augmentations.high_shelf_filter import HighShelfFilter
--- a/audiomentations/augmentations/gain_peak.py
+++ b/audiomentations/augmentations/gain_peak.py
@@ -1,0 +1,41 @@
+import random
+import numpy as np
+
+from audiomentations.core.transforms_interface import BaseWaveformTransform
+
+class GainPeak(BaseWaveformTransform):
+    """
+    Apply a gain peak, i.e turn the gain up, and then down.
+    """
+    
+    supports_multichannel = True
+    
+    def __init__(self, min_gain, min_gain_diff, max_gain_diff, p=0.5):
+        """
+        :param p: The probability of applying this transform
+        """
+        super().__init__(p)
+        self.min_gain = min_gain
+        self.min_gain_diff = min_gain_diff
+        self.max_gain_diff = max_gain_diff
+        
+    def randomize_parameters(self, samples, sample_rate):
+        super().randomize_parameters(samples, sample_rate)
+        if self.parameters["should_apply"]:
+            self.parameters["gain_diff"] = random.uniform(self.min_gain_diff, self.max_gain_diff)
+        
+    def apply(self, samples, sample_rate):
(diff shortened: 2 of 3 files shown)


iver56:
This needs default parameters, so I can do `my_transform = GainPeak()` and start using it just like that

Also, I tried to illustrate some examples of gain curves (green) that I imagine this class could generate and apply in the future (when it is in a more finished state):

bilde

One thing to note from this illustration is that I imagine that it could use an offset, so that the peak can actually appear anywhere, even before the start or after the end of the audio

By the way, I'm curious, what was your motivation for starting to make this class? Are you working on some audio AI application that could benefit from this transform?

atamazian:
Judging my your pics, you suggest using `min_gain = 1` in all cases, since the amplified signal (green) has the same level as original (blue) in the beginning. As for the offset - should it be selected randomly?

As for motivation - I think this class can be beneficial for some audio AI tasks like recognition of badly received signals (for example, signal you received has some louder parts, and some more quiet parts).

iver56:
Yeah, take the gain curve with a grain of salt, especially the min gain part :P I guess min gain isn't the most important feature here, but the difference between min gain and max gain.

iver56:
Yeah, the offset can be selected randomly

atamazian:
OK, I'll modify the code accordingly.

iver56:
Sweet, thanks 👍

atamazian:
As for offset, I think that one can additionally use `Shift` augmentation if necessary. What do you think?

iver56:
Good question! Let's consider the possibilities:

1. `Shift` and then `GainPeak`. In this case the peak will always be centered, so that's maybe not so realistic, and a ML model may learn that any gain peaks always occur in the middle.
2. `GainPeak` and then `Shift`. In this case the gain peak will always be applied to the same part of the input audio (the part that was in the middle before the shift). This is maybe also not so realistic, as in reality it's more like any part of the audio can have a peak.

So in order to offer more variability in the results, I still think it's better that the gain peak itself gets an offset, instead of relying on `Shift`

atamazian:
1. Actually it won't be the case since peak position in `GainPeak` is chosen randomly (it's not always the center)
2. Also won't be the case (see 1)

atamazian:
Can you reply? I'll modify my code more if necessary.

iver56:
Hi :) Thanks for the effort so far, and thanks for the patience. I've been in crunch mode at work for the past few days.

I saw that there are at least these two things that I would like to check before merge:

1. Peak position (offset) - it can be anywhere, even before or after the time of the sound
2. Difference between min gain and max gain

I'll try to give it a look at some point in the coming days

atamazian:
Nice! And thanks for your reply.

iver56:
I had another look at this now, and gathered some thoughts. Here's what I imagine would be good to have in a GainPeak transform:

- [ ] The peak (offset) can be anywhere, even before start or after end of the given input audio, as I illustrated above
- [ ] Various fade curves, not just linear. Fades in "decibel domain" are probably a good idea.
- [ ] Variable fade-in and fade-out durations, independent of each other
- [ ] Gain diff in decibels
- [ ] It should be able to have regions of the sound with constant gain, e.g. before the fade in, at the peak (hold the peak gain for some time) or after the fade out

I imagine that it would also be nice if it could invert its behavior, so it'll essentially be a gain dip instead of a gain peak. Maybe this can be achieved with the gain diff parameter.

At the moment there's quite a gap between what I imagine/desire and what is coded thus far in this pull request. Here's what I propose: I make a GainPeak some time later. When I start doing that, I'll close this PR and make my own branch where I cherry-pick your commits into it, so you'll be listed as a contributor, and then I'll try to implement all the features I suggested above.

atamazian:
OK, let's do as you suggest.

PR #209, opened 2022-06-29
Lambda augmentation for arbitary operations

Thanatoz-1:
With this augmentation the user will be able to apply any required operation over the samples. Some of the popular use cases of this operation could be as follows:
- Polarity inversion
```python
import audiomentations as au

transforms = au.Compose([
    au.Lambda(lambda x: -x, p=1)
])
transforms(np.ones((2, 44100)).astype(np.float32), sample_rate=44100)
```
```bash
>>> array([[-1., -1., -1., ..., -1., -1., -1.],
>>>        [-1., -1., -1., ..., -1., -1., -1.]], dtype=float32)
````

- Clipping
You can use numpy functions such as clipping over the samples and pass the parameters by passing them along with the main function as follows:
```python
transforms = au.Compose([
    au.Lambda(np.clip, p=1, a_min=0.2, a_max=0.8)
])
transforms(np.random.rand(2, 44100).astype(np.float32), sample_rate=44100)
```

- Normalization
```python
transforms = au.Compose([
    au.Lambda(minmax_scale, p=1, copy=1)
])

### CODE REVIEW

**Comment by Thanatoz-1** [comment]
Sorry for the late response, but it took me some time to come up with the use-cases for this augmentation. 
The inspiration for this transform comes from [albumentation's lambda transform](https://albumentations.ai/docs/api_reference/augmentations/transforms/#albumentations.augmentations.transforms.Lambda). This would allow users to have a little more fine-grained control over the operations. I have come up with some potential use cases as follows:

1. __Interchanging the channel or sample positions in the signal__ and also interface with other libraries which might be using the (sample, channel) conventions.
```python
import audiomentations as au
import soundfile as SF

wave, sr = sf.read("./LJ001-0003-2.wav")

transform = au.Compose([
    au.Lambda(lambda x: x.T, p=1),
    ... # Some operations from albumentations
    au.Lambda(lambda x: x.T, p=1),
    ... # Some operations from other libraries that use (sample, channel) convention.

])

```
2. __Apply padding or truncate time-scaled output__
To remove some boundary effects in operations such as Convolutional Reverb, it would be easier to simply add padding inside the `Compose` itself to make the code cleaner. 
 
3. __Apply transformations on individual channels__
In order to apply specific set of augmentation on individual channels, we can create functions to chain the operations and then we can also call them in another `Compose` function. 
```python
transforms1 = au.Compose([au.Gain(p=1)])
transforms2 = au.Compose([au.BandPassFilter(p=1)])

def channel_transforms(x, channel1_transforms, channel2_transforms, sample_rate):
    x = np.stack([channel1_transforms(x[:, 0], sr), channel1_transforms(x[:, 2], sr)])
    return x

transforms_overall = au.Compose([
    ... # Some operations
    au.Lambda(lambda x: x.T, p=1),
    au.Lambda(lambda x: channel_transforms(x, transforms1, transforms2, sr), p=1),
    au.TimeStretch(),
    au.Lambda(lambda x: x.T, p=1)
    ... # Some more operations
])
transformed = transforms_overall(wave, sample_rate=sr)
```
5. __Translate the range of wave signal__
Some networks such as _HiFiGAN vocoder_ need specific operations (such as dividing the entire signal by a specific value) on the signals or some operations such as transforming the signal range between [-1,1] or [0,1]. In order to integrate the augmentations in the pipeline, we will need to translate it back to the original value and then perform the operation. This could also be achieved by using Lambda augmentation having transformations in the compose itself such that the input and output are consistent and the code looks clean.

iver56:
Thanks for the detailed reply! I will consider this after my holiday :) Meanwhile, thanks for your patience and have a good summer

audiomentations/augmentations/fnlambda.py
```diff
@@ -0,0 +1,30 @@
...
+        :param operator: A callable to be applied over samples
+        :param p: The probability of applying this transform
+        :param **kwrags: The parameters and the values to be passed to the operator.
+        """
+        super().__init__(p = p)
+        self.operator = operator
+        self.kwargs = kwargs
+
+    def apply(self, samples, sample_rate):
+        return self.operator(samples, **self.kwargs)
```
iver56, line 30: It should also pass `sample_rate` to make it possible to use it for operations that depend on the sample rate

iver56 (changes requested):

iver56 (approved):

iver56:
Thanks!

audiomentations/augmentations/fnlambda.py
Thanatoz-1: Thanks a lot for merging the pull request. 
I've also updated the examples above to work with the existing changes for reference.

PR #210, opened 2022-07-15
RandomResample

ZFTurbo:
First downsample then upsample back to original shape.


iver56:
What is the motivation for adding this transform? The effect would be pretty much the same as applying a low-pass filter, right?

ZFTurbo:
I don't know if effect is the same. I was trying to emulate users behaviour during video/audio editing.

ZFTurbo:
May be we can add it in repo but add your comment that it's possible to replace with LPF? It's for the persons who will seek for the same augmentation. I don't see a problem to have it in repo.

PR #211, opened 2022-07-29
modify shift and add test

Cangonin:
This PR enables to freeze the Shift transform and then to use it on audio with different sample rates. As it was before, the `num_places_to_shift` would still be the same after freezing the parameters no matter the sample rate.

### DIFF
--- a/audiomentations/augmentations/shift.py
+++ b/audiomentations/augmentations/shift.py
@@ -49,15 +49,14 @@ class Shift(BaseWaveformTransform):
     def randomize_parameters(self, samples, sample_rate):
         super().randomize_parameters(samples, sample_rate)
         if self.parameters["should_apply"]:
-            self.parameters["num_places_to_shift"] = int(
-                round(
-                    random.uniform(self.min_fraction, self.max_fraction)
-                    * samples.shape[-1]
-                )
+            self.parameters["fraction_to_shift"] = random.uniform(
+                self.min_fraction, self.max_fraction
             )
 
     def apply(self, samples, sample_rate):
-        num_places_to_shift = self.parameters["num_places_to_shift"]
+        num_places_to_shift = round(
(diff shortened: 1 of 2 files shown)


Cangonin:
The changes I did seem to be  breaking changes. Maybe I could add `fraction_to_shift` as a second parameter ? And deduce `num_places_to_shift` from `fraction_to_shift` in `apply`

audiomentations/augmentations/shift.py
```diff
@@ -49,15 +49,14 @@ def __init__(
...
-                    random.uniform(self.min_fraction, self.max_fraction)
-                    * samples.shape[-1]
-                )
+            self.parameters["fraction_to_shift"] = random.uniform(
+                self.min_fraction, self.max_fraction
             )
 
     def apply(self, samples, sample_rate):
-        num_places_to_shift = self.parameters["num_places_to_shift"]
+        num_places_to_shift = round(
```
iver56, line 57: int?

iver56 (commented):
Thanks! I'll have a look at this on wednesday :)

audiomentations/augmentations/shift.py
Cangonin: If the argument `ndigits ` in the function `round ` is omitted, it returns an `int` already, that's why I removed the int()

iver56:
> The changes I did seem to be breaking changes

Why was it a breaking change?

Cangonin:
> Why was it a breaking change?

Because I replaced the parameter `num_places_to_shift` by `fraction_to_shift`. Or maybe this is not a breaking change ?

iver56:
I guess it's only breaking for those who inspect the parameters. That should be only a few, as far as I know. I allow this change, it doesn't feel too breaking.

PR #219, opened 2022-09-01
Update shift.py

huangxu1991:
when shift to left, the rightmost num_places_to_shift need to be zero, so it should be negative num_places_to_shift

### DIFF
--- a/audiomentations/augmentations/shift.py
+++ b/audiomentations/augmentations/shift.py
@@ -63,7 +63,7 @@ class Shift(BaseWaveformTransform):
             if num_places_to_shift > 0:
                 shifted_samples[..., :num_places_to_shift] = 0.0
             elif num_places_to_shift < 0:
-                shifted_samples[..., num_places_to_shift:] = 0.0
+                shifted_samples[..., -num_places_to_shift:] = 0.0
 
         if self.fade:
             fade_length = int(sample_rate * self.fade_duration)


mmxgn:
Hi.

I don't think that's right.

`num_places_to_shift` is given as a negative number, if you add a negation there its becoming a positive number and then all elements above than number get zeroed.

iver56:
The case you are refering to is tested here: https://github.com/iver56/audiomentations/blob/master/tests/test_shift.py#L51

If you think the test looks correct, I think we can close this PR

huangxu1991:
for example, for array [1, 2, 3, 4, 5, 6] and not rollover,
=> shift 2 steps towards right, becomes => [5, 6, 1, 2, 3, 4], remove leftmost 2 =>[0, 0, 1, 2, 3, 4]
=> shift 2 steps towards left,    becomes => [3, 4, 5, 6, 1, 2], in fact we want [3, 4, 5, 6, 0, 0], but as the code
` shifted_samples[..., num_places_to_shift:] = 0.0`
it becomes [3, 4, 0, 0, 0, 0], it's absoultly wrong! so it should be like,
` shifted_samples[..., -num_places_to_shift:] = 0.0`

iver56:
Did you look at the unit test I referred to? And in your opinion, do you think it looks correct or incorrect?

huangxu1991:
> Did you look at the unit test I referred to? And in your opinion, do you think it looks correct or incorrect?

sorry, I just test it! It's right in fact, so embrassing, i'm too careless (I even ignore the negative if condition)
the pr can be closed!

Issue #224, opened 2022-09-12
Change default noise_rms="relative_to_whole_input" in AddShortNoises?


iver56:
The default value for noise_rms is currently "relative", but "relative_to_whole_input" might actually be a better default, given that the loudness of the given signal can vary from very silent to very loud, and "relative" doesn't allow placing any loud short noises in part that is silent in the input

What do you think @Cangonin?

Cangonin:
I agree!

atamazian:
This issue is solved, so I suggest to close it.

iver56:
I can close this issue after we make the final switch. For now it just says this:

```
            warnings.warn(
                "The default value of noise_rms in AddShortNoises will change from"
                " 'relative' to 'relative_to_whole_input' in a future version of"
                " audiomentations. If you set the value explicitly, you will not get"
                " this warning.",
                FutureWarning
            )
```

It would be good to make the switch in the next version 👍

PR #225, opened 2022-09-12
Change default of noise_rms in add short noise transform

Cangonin:
Implements https://github.com/iver56/audiomentations/issues/224.


iver56:
I would prefer to keep the old default value (with warning) for now, for backwards compatibility. We can make the final switch some time later. Just like this:

https://github.com/iver56/audiomentations/commit/143f9c2b0ca002563a25f3ca47e9653ec32d2c24

PR #244, opened 2022-11-02
add randomized shift

lianabagh:


iver56:
Could you write a few sentences that explains what this transform does, what it sounds like and when it is useful?

Issue #241, opened 2022-10-06
Implement AdjustDuration transform


iver56:
Should have a target duration parameter that can be specified either in seconds or in number of samples

If the input sound is longer than the target duration, pick a random offset (so we don't always output just the beginning of the audio) and crop the sound to the target duration

If the input sound is shorter than the target duration, pad the end of the sound (append digital silence) so the duration matches the target duration. Maybe it makes sense to support various padding modes here, just like in the `Padding` transform.

DaGr8Nave:
has this been implemented?

iver56:
No, not yet. If you are interested in helping, a pull request would be welcome 👍

PR #270, opened 2023-03-18
Add DurationClip

BakerBunker:


iver56:
Thanks for the PR! Tagging this related issue: https://github.com/iver56/audiomentations/issues/241

audiomentations/augmentations/duration_clip.py
```diff
@@ -0,0 +1,74 @@
...
+        self,
+        duration_samples: int = None,
+        duration_seconds: float = None,
+        pad_mode: str = "constant",
+        pad_section: str = "end",
+        p: float = 0.5,
+    ):
+        """
+        :param duration_samples: Target duratin in sample length
+        :param duration_seconds: Target duratin in second
```
iver56, line 27:
```suggestion
        :param duration_samples: Target duration in number of samples
        :param duration_seconds: Target duration in seconds
```

audiomentations/augmentations/duration_clip.py
```diff
@@ -0,0 +1,74 @@
+import numpy as np
+
+from audiomentations.core.transforms_interface import BaseWaveformTransform
+
+
+class DurationClip(BaseWaveformTransform):
```
iver56, line 6: I would prefer RandomCrop over DurationClip, because the term Clip is can confused with the [`Clip` transform](https://iver56.github.io/audiomentations/waveform_transforms/clip/) that we already have, which relates to clipping distortion. A clip can also refer to a small segment of audio that is extracted from a larger file.

I'm not 100% sure about the name RandomCrop though, because cropping is commonly used about images/rasters (e.g. [like this in albumentations](https://albumentations.ai/docs/api_reference/augmentations/crops/transforms/)), but I guess it _could_ sometimes be used about audio too. Some industry people like to say trim or cut. Maybe CutToLength could work? Or TrimPadToLength? 🤔 @Cangonin @mmxgn Do you have any thoughts/opinions on this?

```suggestion
class RandomCrop(BaseWaveformTransform):
```

audiomentations/augmentations/duration_clip.py
```diff
@@ -0,0 +1,74 @@
...
+        if duration_samples is not None and duration_seconds is not None:
+            raise ValueError(
+                "should have duration_samples or duration_seconds, but not both"
+            )
+        elif duration_seconds:
+            assert duration_seconds > 0
+            self.get_target_samples = lambda sr: int(duration_seconds * sr)
+        elif duration_samples:
+            assert duration_samples > 0
+            self.get_target_samples = lambda sr: duration_samples
```
iver56, line 49: Raise an exception if both duration_samples _and_ duration_seconds are given. Only one of them should be given.

iver56 (changes requested):

audiomentations/augmentations/duration_clip.py
BakerBunker: "both are given" exception is raised in line 40.
"none of them are given" exception is an assertion, and it at line 39.
Now the expression is indeed relatively unclear，I change it in another suggestion.

audiomentations/augmentations/duration_clip.py
mmxgn: Isn't this functionally the same as a TimeTrim transform just implemented differently?

iver56: I don't know. Is that an established term?
bilde

mmxgn: How about `AudioLengthAdjust`?

iver56: Or AdjustAudioLength? Or AdjustLength or LengthAdjust, since all audiomentations transforms work on audio anyway? 🤷

Naming sure is hard xD

Cangonin: I agree that clipping refers more to distortion, that's how I interpreted it when I saw the name of this transform. Since the audio can be trimmed or padded after this transform, I like your suggestion `AdjustDuration` :)

mmxgn: I also like `AdjustDuration` too.

iver56: Does it sound like "adjusting the duration" is a time stretching operation though? 🤔

Cangonin: I see your point with `AdjustDuration`. I think that `AdjustLength` works too.

iver56: AI gives a vote to AdjustDuration

bilde

Cangonin: What about `Resize` ?

iver56: The term `Resize` is common in image processing, but that implies some sort of resampling, which is not what we are doing here. I think we'll roll with AdjustDuration, since that seems to have the most votes now

Issue #264, opened 2023-02-27
Add BitCrush


mmxgn:

iver56:
Yup. That page mainly features two different transforms:

1. Sample rate reduction without low-pass filtering. Also known as sample-and-hold or decimation, but I personally like to think of it as nearest neighbour. This is a terrible form of resampling audio quality wise, but it leads to an interesting kind of output that can be used in music (often electronic music), and it can be used for data augmentation if not overdone
2. Bit depth reduction, which reduces the number of bits that can be used for representing each audio sample. This introduces quantization noise which can mess with dynamic range and harmonics, amongst other things.

I think these two are sufficiently different that they should be different augmentation classes in audiomentations.

Based on the issue title, I think this one mainly refers to 2. Bit depth reduction

Would you like to implement it? A PR would be welcome

iver56:
Here's another nice resource on this topic, from a course I took at NTNU a long time ago: http://gdsp.hf.ntnu.no/lessons/1/4/

atamazian:
Thanks for the info. I'll implement this. Please assign this issue to me.

iver56:
TODO:

- [ ] Add documentation page
- [ ] Link to it from readme

PR #319, opened 2024-03-02
Add BitCrush

atamazian:
Attempt to implement https://github.com/iver56/audiomentations/issues/264.

### DIFF
--- a/audiomentations/augmentations/bit_crush.py
+++ b/audiomentations/augmentations/bit_crush.py
@@ -1,0 +1,26 @@
+import numpy as np
+from numpy.typing import NDArray
+
+from audiomentations.core.transforms_interface import BaseWaveformTransform
+
+class BitCrush(BaseWaveformTransform):
+    """
+    Apply a bit crush effect the audio by reducing of the resolution or bandwidth of audio 
+    data. This class implements bit depth reduction, which reduces the number of bits 
+    that can be used for representing each audio sample.
+    """
+    
+    supports_multichannel = True
+    
+    def __init__(self, bit_depth: int = 4, p: float = 0.5):
+        """
+        :param bit_depth: The bit depth to which the audio sample will be converted to
+        :param p: The probability of applying this transform
+        """
+        super().__init__(p)
+        self.bit_depth = bit_depth
+        
+    def apply(self, samples: NDArray[np.float32], sample_rate: int):
+        q = (2**self.bit_depth/2) + 1
+        return np.round(samples * q)/q
+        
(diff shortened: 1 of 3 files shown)


audiomentations/augmentations/bit_crush.py
```diff
@@ -0,0 +1,26 @@
...
+class BitCrush(BaseWaveformTransform):
+    """
+    Apply a bit crush effect the audio by reducing of the resolution or bandwidth of audio 
+    data. This class implements bit depth reduction, which reduces the number of bits 
+    that can be used for representing each audio sample.
+    """
+    
+    supports_multichannel = True
+    
+    def __init__(self, bit_depth: int = 4, p: float = 0.5):
```
iver56, line 15: In the context of data augmentation, it would be nice if the class can randomize the bit depth parameter, like this:
```suggestion
    def __init__(self, min_bit_depth: int = 5, max_bit_depth: int = 10, p: float = 0.5):
```

Can you add this functionality?

iver56 (changes requested):

audiomentations/augmentations/bit_crush.py
atamazian: Sure, I'll do it.

atamazian: Done

audiomentations/augmentations/bit_crush.py
```diff
@@ -0,0 +1,38 @@
...
+        self.max_bit_depth = max_bit_depth
+        
+        assert self.min_bit_depth <= self.max_bit_depth
+        
+    def randomize_parameters(self, samples: NDArray[np.float32], sample_rate: int):
+        super().randomize_parameters(samples, sample_rate)
+        if self.parameters["should_apply"]:
+            self.parameters["bit_depth"] = random.uniform(
+                self.min_bit_depth, self.max_bit_depth
+            )
```
iver56, line 33: This becomes a float, and not an int. Was this intentional?

atamazian: No, that's my error.

audiomentations/augmentations/bit_crush.py
```diff
@@ -28,8 +28,10 @@ def __init__(self, min_bit_depth: int = 5, max_bit_depth: int = 10, p: float = 0
     def randomize_parameters(self, samples: NDArray[np.float32], sample_rate: int):
         super().randomize_parameters(samples, sample_rate)
         if self.parameters["should_apply"]:
-            self.parameters["bit_depth"] = random.uniform(
-                self.min_bit_depth, self.max_bit_depth
+            self.parameters["bit_depth"] = int(
+                random.uniform(
+                    self.min_bit_depth, self.max_bit_depth
+                )
             )
```
iver56, line 35: using random.randint() without an int() around it would be even better, because that would include the max_bit_depth

atamazian: Nice suggestion. Done!

iver56 (approved):
Thanks! I might refine it a little bit (e.g. add docs) before it gets released later

Issue #261, opened 2023-02-24
Feature Request: Aliasing


mmxgn:
This is a nice degradation which would help imho in restoration algorithms and is provided in the Audio Degradation Toolbox. E.g. do the following:

1. Naively downsample to a lower sample rate without filtering
2. Upsample normally.

iver56:
Yep, good idea

atamazian:
Can you assign me to this issue?

iver56:
Ah, yes, here we have the issue for the stuff I mentioned in #264 

> Sample rate reduction without low-pass filtering. Also known as sample-and-hold or decimation, but I personally like to think of it as nearest neighbour. This is a terrible form of resampling audio quality wise, but it leads to an interesting kind of output that can be used in music (often electronic music), and it can be used for data augmentation if not overdone
> http://gdsp.hf.ntnu.no/lessons/1/3/

I will assign the issue to you, atamazian 👍 Thanks for contributing

iver56:
TODO:
- [ ] Add documentation page
- [ ] Link to it from readme

PR #322, opened 2024-03-05
Add Aliasing

atamazian:
Attempt to implement https://github.com/iver56/audiomentations/issues/261.

### DIFF
--- a/audiomentations/augmentations/aliasing.py
+++ b/audiomentations/augmentations/aliasing.py
@@ -4,6 +4,7 @@ import numpy as np
 from numpy.typing import NDArray
 
 from audiomentations.core.transforms_interface import BaseWaveformTransform
+from audiomentations.core.utils import convert_frequency_to_mel, convert_mel_to_frequency
 
 
 class Aliasing(BaseWaveformTransform):
@@ -25,14 +26,14 @@ class Aliasing(BaseWaveformTransform):
         if min_sample_rate > max_sample_rate:
             raise ValueError("min_sample_rate must not be larger than max_sample_rate")
         
-        self.min_sample_rate = min_sample_rate
-        self.max_sample_rate = max_sample_rate
+        self.min_mel = convert_frequency_to_mel(min_sample_rate)
+        self.max_mel = convert_frequency_to_mel(max_sample_rate)
 
     def randomize_parameters(self, samples: NDArray[np.float32], sample_rate: int):
         super().randomize_parameters(samples, sample_rate)
         if self.parameters["should_apply"]:
-            self.parameters["new_sample_rate"] = random.randint(
-                self.min_sample_rate, self.max_sample_rate
+            self.parameters["new_sample_rate"] = convert_mel_to_frequency(
+                random.uniform(self.min_mel, self.max_mel)
             )
 
     def apply(self, samples: NDArray[np.float32], sample_rate: int):
--- a/audiomentations/augmentations/aliasing.py
+++ b/audiomentations/augmentations/aliasing.py
@@ -1,0 +1,43 @@
+import random
+
+import numpy as np
+from numpy.typing import NDArray
+from scipy import signal
+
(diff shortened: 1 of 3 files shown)


tests/test_aliasing.py
```diff
@@ -0,0 +1,30 @@
...
+        num_channels = 3
+        samples = np.random.normal(0, 0.1, size=(num_channels, 2048)).astype(np.float32)
+        sample_rate = 16000
+        augmenter = Aliasing(min_sample_rate=8000, max_sample_rate=32000, p=1.0)
+
+        distorted_samples = augmenter(samples=samples, sample_rate=sample_rate)
+
+        assert samples.dtype == distorted_samples.dtype
+        assert samples.shape == distorted_samples.shape
+        assert len(distorted_samples) == len(samples)
```
iver56, line 29: This is line redundant, as the assert in the line above already compares the shape with all the dimensions

audiomentations/augmentations/aliasing.py
```diff
@@ -0,0 +1,43 @@
...
+        if self.parameters["should_apply"]:
+            self.parameters["new_sample_rate"] = random.randint(
+                self.min_sample_rate, self.max_sample_rate
+            )
+
+    def apply(self, samples: NDArray[np.float32], sample_rate: int):
+        n = len(samples)
+        dwn_n = round(n * float(self.parameters["new_sample_rate"]) / sample_rate)
+        dwn_samples = signal.resample(samples, dwn_n)
+        return signal.resample(dwn_samples, n)
```
iver56, line 43: Using this resampling method does produce aliasing artifacts, which is what we want to achieve in this transform. We need something like nearest neighbour

iver56 (changes requested):

audiomentations/augmentations/aliasing.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def randomize_parameters(self, samples: NDArray[np.float32], sample_rate: int):
+        super().randomize_parameters(samples, sample_rate)
+        if self.parameters["should_apply"]:
+            self.parameters["new_sample_rate"] = random.randint(
+                self.min_sample_rate, self.max_sample_rate
+            )
+
+    def apply(self, samples: NDArray[np.float32], sample_rate: int):
+        n = len(samples)
```
iver56, line 40: Using len won't work correctly when dealing with stereo audio, e.g. shapes like (2, 86400)

This alternative works for both mono audio and multi-dimensional arrays:

```suggestion
        n = samples.shape[-1]
```

audiomentations/augmentations/aliasing.py
atamazian: You probably meant "does not produce". OK, I'll change this to NN.

audiomentations/augmentations/aliasing.py
atamazian: OK, will fix

tests/test_aliasing.py
atamazian: OK, will fix

audiomentations/augmentations/aliasing.py
iver56: Yes, you are right. I have now corrected my message

atamazian: Is it correct now? I used `np.interp` for downsampling/upsampling, and the resulting array is no longer equal to the original one.

iver56: Yes, this way of doing it will indeed lead to aliasing artifacts, even though interpolation is not the same thing as nearest neighbour (also known as sample and hold), which will sound even harsher

atamazian: I see. Should I keep `np.interp` solution, or exactly NN is needed?

atamazian:
I get it, will do.

PR #336, opened 2024-04-26
Don't calculate heavy room parameters if augmentation is not going to be applied

qwaker00:
For small `p` it will save a lot of time.


iver56:
Great! Thanks for your contribution

Issue #285, opened 2023-06-22
add `__repr__` and `__str__` methods for `Compose` etc


aaprasad:
`albumentations` has a nice `__repr__` and `__str__` for their `Compose` class that shows the augmentations being used. Would be nice to have the same thing in Compose instead of the generic `<audiomentations.core.composition.Compose object at 0x15546be0cdf0>`

atamazian:
Can you assign this issue to me?

atamazian:
I can tell that it's definitely not "a good first issue". I've made it work though, will commit soon.

PR #361, opened 2024-11-02
Add __repr__ similar to albumentations

atamazian:
Attempt to implement https://github.com/iver56/audiomentations/issues/285.

### DIFF
--- a/audiomentations/augmentations/add_background_noise.py
+++ b/audiomentations/augmentations/add_background_noise.py
@@ -178,14 +178,3 @@ class AddBackgroundNoise(BaseWaveformTransform):
         )
         del state["_load_sound"]
         return state
-        
-    def get_transform_init_args_names(self) -> tuple[str, ...]:
-        return (
-            "sound_file_paths",
-            "min_snr_db",
-            "max_snr_db",
-            "noise_rms",
-            "min_absolute_rms_db",
-            "max_absolute_rms_db",
-            "noise_transform"
-        )
--- a/audiomentations/augmentations/add_background_noise.py
+++ b/audiomentations/augmentations/add_background_noise.py
@@ -178,3 +178,14 @@ class AddBackgroundNoise(BaseWaveformTransform):
         )
         del state["_load_sound"]
         return state
+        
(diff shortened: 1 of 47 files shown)


atamazian:
I see your point, will try to implement.

atamazian:
I was simply following `albumentations` syntax.

atamazian:
I've used `inspect` to replace `get_transform_init_args_names`. Is this better? As for the test, I used a simple script to check this. When you approve final `__repr__` realization, I'll write an "official" test for this.

iver56:
Yeah, that's better. You went from adding 500 lines to adding just 200 lines. That feels more manageable.

I gave the repr a try, and I like the style in which it represents objects. You can start writing an "official" test for this.

During my testing, I uncovered an issue. What I did was I added a `print(augmenter)` in the loop in demo/demo.py and tried to run it. I ran into multiple exceptions, for example:

'AddBackgroundNoise' object has no attribute 'sounds_path'
'ApplyImpulseResponse' object has no attribute 'ir_path'
'AddShortNoises' object has no attribute 'sounds_path'
'AdjustDuration' object has no attribute 'duration_samples'
'Aliasing' object has no attribute 'min_sample_rate'
'SevenBandParametricEQ' object has no attribute 'min_gain_db'

The problem seems to be that the __repr__ tries to access attributes that do not exist, resulting in an exception. Apparently, our current test suite did not catch all these issues.

I guess we need to
1) add missing attributes
2) tolerate missing attribute instead of throwing an exception - this would feel a little bit safer in case we miss some cases

atamazian:
I'm glad to hear you liked it! I'll modify the code accordingly.

atamazian:
I've made some modifications to the code. Is it better now?

iver56:
Yes

atamazian:
I'm not sure how a test should look like for `__repr__`. Should I compare two strings? Or where is a better way of doing it?

iver56:
Something like

```
transform = Compose([GainTransition(), Shift()])
assert str(transform) == "whatever it is supposed to be"
```

iver56:
Neat! Thanks for the contribution

Issue #312, opened 2024-02-06
Add Chorus transform


iver56:

atamazian:
Can you elaborate a little?

iver56:
Explanation and audio example here: http://gdsp.hf.ntnu.no/lessons/14/53/

atamazian:
Can you assign this issue to me?

PR #363, opened 2024-11-16
Add Chorus transform

atamazian:
Attempt to implement https://github.com/iver56/audiomentations/issues/312.


iver56:
Thanks for the PR, and thanks for the patience. I will have a look at it next week

PR #387, opened 2025-04-20
Add optional weights argument to OneOf for weighted random selection

worldveil:
## Description

This addition should leave all existing scripts working the same, but allow for an optional argument for `OneOf` that allows weighted selection of the transform, instead of uniform random. 

If `weights` are not supplied, behavior will default to the way it has always been (uniform random), so this is fully backwards compatible :)

## Motivation

I have a large number of impulse response (IR) files that I wish to use at random in some augmentations for a model I am training, for example from my repo: [impulse-response-audio-files](https://github.com/worldveil/impulse-response-audio-files).

However, not all categories of these IRs are equally as important to our usecase, and as such, we'd like to upsample certain classes (ie: venues, large rooms), and downsample others (ie: small rooms, misc other IRs). 

## Example script

Let's envision a scenario in which we'd like to upsample one category 4x vs another (0.8 vs 0.2):

```python
from pathlib import Path
from collections import Counter

from audiomentations import Compose, ApplyImpulseResponse, OneOf
import numpy as np

# load some impulse responses from: https://github.com/worldveil/impulse-response-audio-files
IR_DIR = "./impulse-response-audio-files"
venues_dir1 = Path(IR_DIR) / "echo_thief" / "Venues"
misc_dir2 = Path(IR_DIR) / "voxengo"
weights = [0.8, 0.2]

augment = Compose([
    OneOf([
        ApplyImpulseResponse(ir_path=venues_dir1, p=1.0),
        ApplyImpulseResponse(ir_path=misc_dir2, p=1.0)
    ], p=0.5, weights=weights),
])

# Generate 2 seconds of dummy audio for the sake of example
samples = np.random.uniform(low=-0.2, high=0.2, size=(32000,)).astype(np.float32)

# let's count the number of times each transform is chosen to verify it's working
index_chosen = Counter()
one_of = augment.transforms[-1]

for i in range(1000):
    # Augment/transform/perturb the audio data
    augmented_samples = augment(samples=samples, sample_rate=44100)

    # roll the dice
    for i, transform in enumerate(augment.transforms):
        if i == len(augment.transforms) - 1:
            #print(f"We chose transform: {one_of.transforms[transform.transform_index].__class__}, which was index={transform.transform_index} (weights={one_of.weights})")
            index_chosen[transform.transform_index] += 1

# we should expect to see zero 4x as many times as 1
print(index_chosen)
```

For me, the output is:

```
Counter({0: 791, 1: 209})
```

which aligns with our expectation.

## Tests

Are all passing, with the exception of the two that depend on `fast-align-audio` which I wasn't able to install in the 30 min I had to make this PR :) but from quick glance this PR should not affect that functionality at all.

### DIFF
--- a/audiomentations/core/composition.py
+++ b/audiomentations/core/composition.py
@@ -7,6 +7,7 @@ from numpy.typing import NDArray
 from audiomentations.core.transforms_interface import BaseSpectrogramTransform
 from audiomentations.core.serialization import get_shortest_class_fullname
 from audiomentations.core.utils import format_args
+from audiomentations.core.sampling import WeightedChoiceSampler
 
 REPR_INDENT_STEP = 2
 
@@ -278,7 +279,7 @@ class OneOf(BaseCompose):
         super().__init__(transforms, p)
         self.transform_index = 0
         self.should_apply = True
-        
+
         # Handle weights initialization and validation
         if weights is not None:
             if len(weights) != len(transforms):
@@ -292,14 +293,15 @@ class OneOf(BaseCompose):
             self.weights /= np.sum(self.weights)
         else:
             # If no weights provided, use uniform distribution
-            # np.random.choice() will use the uniform distribution if p=None
-            self.weights = None
+            self.weights = np.ones(len(transforms)) / len(transforms)
+
+        self.sampler = WeightedChoiceSampler(self.weights)
 
     def randomize_parameters(self, *args, **kwargs):
         super().randomize_parameters(*args, **kwargs)
         self.should_apply = random.random() < self.p
         if self.should_apply:
-            self.transform_index = np.random.choice(len(self.transforms), p=self.weights)
+            self.transform_index = self.sampler.sample(size=1)[0]
 
     def __call__(self, *args, **kwargs):
         if not self.are_parameters_frozen:
--- a/audiomentations/core/composition.py
+++ b/audiomentations/core/composition.py
@@ -1,5 +1,5 @@
 import random
-from typing import Any
+from typing import Any, Optional
 
 import numpy as np
 from numpy.typing import NDArray
@@ -265,7 +265,7 @@ class OneOf(BaseCompose):
     augment = OneOf([
         TimeStretch(min_rate=0.8, max_rate=1.25, p=1.0),
         PitchShift(min_semitones=-4, max_semitones=4, p=1.0),
-    ], weights=[0.8, 0.2])
+    ], weights=[0.2, 0.8])
     # Generate 2 seconds of dummy audio for the sake of example
     samples = np.random.uniform(low=-0.2, high=0.2, size=(32000,)).astype(np.float32)
     # Augment/transform/perturb the audio data
@@ -274,7 +274,7 @@ class OneOf(BaseCompose):
     ```
     """
 
-    def __init__(self, transforms, p: float = 1.0, weights: list[float] | None = None):
+    def __init__(self, transforms, p: float = 1.0, weights: Optional[list[float]] = None):
         super().__init__(transforms, p)
         self.transform_index = 0
         self.should_apply = True
@@ -288,10 +288,12 @@ class OneOf(BaseCompose):
             if sum(weights) == 0:
                 raise ValueError("Sum of weights must be greater than 0")
             # Normalize weights to sum to 1
-            self.weights = np.array(weights) / sum(weights)
+            self.weights = np.array(weights)
+            self.weights /= np.sum(self.weights)
         else:
             # If no weights provided, use uniform distribution
-            self.weights = np.ones(len(transforms)) / len(transforms)
+            # np.random.choice() will use the uniform distribution if p=None
+            self.weights = None
 
     def randomize_parameters(self, *args, **kwargs):
         super().randomize_parameters(*args, **kwargs)
--- a/audiomentations/core/composition.py
+++ b/audiomentations/core/composition.py
@@ -259,31 +259,45 @@ class SomeOf(BaseCompose):
 class OneOf(BaseCompose):
     """
     OneOf randomly picks one of the given transforms when called, and applies that
-    transform.
+    transform. Optional `weights` can be supplied to guide the probability of each transform being chosen.
     Usage example:
     ```
     augment = OneOf([
         TimeStretch(min_rate=0.8, max_rate=1.25, p=1.0),
         PitchShift(min_semitones=-4, max_semitones=4, p=1.0),
-    ])
+    ], weights=[0.8, 0.2])
     # Generate 2 seconds of dummy audio for the sake of example
     samples = np.random.uniform(low=-0.2, high=0.2, size=(32000,)).astype(np.float32)
     # Augment/transform/perturb the audio data
     augmented_samples = augment(samples=samples, sample_rate=16000)
-    # Result: The audio was either time-stretched or pitch-shifted, but not both
+    # Result: The audio was either time-stretched (less likely) or pitch-shifted (4x more likely), but not both
     ```
     """
 
-    def __init__(self, transforms, p: float = 1.0):
+    def __init__(self, transforms, p: float = 1.0, weights: list[float] | None = None):
         super().__init__(transforms, p)
         self.transform_index = 0
         self.should_apply = True
+        
+        # Handle weights initialization and validation
+        if weights is not None:
+            if len(weights) != len(transforms):
+                raise ValueError("Length of weights must match length of transforms")
+            if any(w < 0 for w in weights):
+                raise ValueError("All weights must be non-negative")
+            if sum(weights) == 0:
+                raise ValueError("Sum of weights must be greater than 0")
+            # Normalize weights to sum to 1
+            self.weights = np.array(weights) / sum(weights)
+        else:
+            # If no weights provided, use uniform distribution
+            self.weights = np.ones(len(transforms)) / len(transforms)
 
     def randomize_parameters(self, *args, **kwargs):
         super().randomize_parameters(*args, **kwargs)
         self.should_apply = random.random() < self.p
         if self.should_apply:
-            self.transform_index = random.randint(0, len(self.transforms) - 1)
+            self.transform_index = np.random.choice(len(self.transforms), p=self.weights)
 
     def __call__(self, *args, **kwargs):
         if not self.are_parameters_frozen:
--- a/audiomentations/core/sampling.py
+++ b/audiomentations/core/sampling.py
@@ -1,0 +1,30 @@
+import numpy as np
+
+
+class WeightedChoiceSampler:
+    """
+    A class for sampling from a weighted distribution using the cumulative distribution function (CDF).
+
+    Usage:
+    >>> sampler = WeightedChoiceSampler(weights=[0.2, 0.8])
+    >>> samples = sampler.sample(size=10)
+    """
+    def __init__(self, weights: list[float] = None):
+        assert weights is not None and len(weights) > 0, "Weights must be provided"
+
+        # ensure weights are normalized
+        weights = np.array(weights)
+        weights /= np.sum(weights)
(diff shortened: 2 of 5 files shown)


audiomentations/core/composition.py
```diff
@@ -259,31 +259,45 @@ def __call__(self, *args, **kwargs):
...
+        # Handle weights initialization and validation
+        if weights is not None:
+            if len(weights) != len(transforms):
+                raise ValueError("Length of weights must match length of transforms")
+            if any(w < 0 for w in weights):
+                raise ValueError("All weights must be non-negative")
+            if sum(weights) == 0:
+                raise ValueError("Sum of weights must be greater than 0")
+            # Normalize weights to sum to 1
+            self.weights = np.array(weights) / sum(weights)
```
iver56, line 291: This should be faster, given that the sum is calculated with numpy instead of python

```suggestion
            self.weights = np.array(weights)
            self.weights /= np.sum(self.weights)
```

audiomentations/core/composition.py
```diff
@@ -259,31 +259,45 @@ def __call__(self, *args, **kwargs):
...
     samples = np.random.uniform(low=-0.2, high=0.2, size=(32000,)).astype(np.float32)
     # Augment/transform/perturb the audio data
     augmented_samples = augment(samples=samples, sample_rate=16000)
-    # Result: The audio was either time-stretched or pitch-shifted, but not both
+    # Result: The audio was either time-stretched (less likely) or pitch-shifted (4x more likely), but not both
     ```
     """
 
-    def __init__(self, transforms, p: float = 1.0):
+    def __init__(self, transforms, p: float = 1.0, weights: list[float] | None = None):
```
iver56, line 277: For Python 3.9 compatibility (we still support that version), we need to do this instead:

```suggestion
    def __init__(self, transforms, p: float = 1.0, weights: Optional[list[float]] = None):
```

audiomentations/core/composition.py
```diff
@@ -259,31 +259,45 @@ def __call__(self, *args, **kwargs):
...
+                raise ValueError("Length of weights must match length of transforms")
+            if any(w < 0 for w in weights):
+                raise ValueError("All weights must be non-negative")
+            if sum(weights) == 0:
+                raise ValueError("Sum of weights must be greater than 0")
+            # Normalize weights to sum to 1
+            self.weights = np.array(weights) / sum(weights)
+        else:
+            # If no weights provided, use uniform distribution
+            self.weights = np.ones(len(transforms)) / len(transforms)
```
iver56, line 294: This avoids dividing every element in the array, which makes it significantly faster

```suggestion
            n = len(transforms)
            self.weights = np.full(n, 1.0 / n, dtype=float)
```

iver56 (changes requested):
Thanks for implementing this! I have actually been thinking about this feature myself lately. Could you please add unit test for this functionality too? I would be grateful for that. Check the code change suggestions too.

audiomentations/core/composition.py
iver56, line 294: Actually, scratch that. Here is a simpler alternative that also uses less memory. This works because the `p` argument in np.random.choice is optional, and `p=None` leads to a uniform distribution.

```suggestion
            self.weights = None
```

iver56:
One thing I was thinking about earlier when considering this feature was the algorithmic complexity, which is important if there are many choices. I do not think there will be typically be lots of choices in OneOf, but I'm thinking of adding this type of feature to other parts of audiomentations too.

Current implementation:
_Call np.random.choice() for every one-off draw._
Memory (the list of weights): O(n)
Time (each one-off draw): O(n)

Alternative implementation:
_Pre-computed CDF in init. Use binary search for each one-off draw._
Memory (the CDF): O(n)
Time (each one-off draw): O(log n)
Con: CDF has to be recomputed if the weights change after init. However, I think it is safe to assume that the weights are static, so this con is not important to consider.

If you're feeling adventurous, please implement the suggested alternative implementation. Then we can reuse that pattern when also adding weights to other parts of audiomentations.

audiomentations/core/composition.py
```diff
@@ -259,31 +259,45 @@ def __call__(self, *args, **kwargs):
...
         TimeStretch(min_rate=0.8, max_rate=1.25, p=1.0),
         PitchShift(min_semitones=-4, max_semitones=4, p=1.0),
-    ])
+    ], weights=[0.8, 0.2])
     # Generate 2 seconds of dummy audio for the sake of example
     samples = np.random.uniform(low=-0.2, high=0.2, size=(32000,)).astype(np.float32)
     # Augment/transform/perturb the audio data
     augmented_samples = augment(samples=samples, sample_rate=16000)
-    # Result: The audio was either time-stretched or pitch-shifted, but not both
+    # Result: The audio was either time-stretched (less likely) or pitch-shifted (4x more likely), but not both
```
iver56, line 273: Could you please double-check the description? At first glance, it seems like it doesn't match the code example above (`[0.8, 0.2]`)

worldveil: Ah! You're right I got them mixed around. I changed the weights above, so they match. Good catch :)

worldveil:
@iver56 ha! I don't think it's as slow as you think:

```python
In [76]: %time np.random.choice(np.random.rand(100_000))
CPU times: user 611 μs, sys: 650 μs, total: 1.26 ms
Wall time: 738 μs
Out[76]: np.float64(0.8334054792762154)
```

this is roughly an order of magnitude faster than reading an audio file from disk. and 100,000 different augmentations would really be a crazy pipeline...

but I went ahead and implemented it in case some other fun ideas get implemented here! 

added a new `sampling.py` along with some unit tests.

iver56:
Thanks for checking the execution time! The unweighted case (without passing `p`) is indeed really fast, even for large sizes (O(1) in theory). I was mainly concerned about the weighted case, which scales linearly (O(n)). Thanks for addressing the PR comments, and for making the WeightedChoiceSampler. I will have another look

worldveil:
Oh, no of course. I copied the wrong thing - duh!

## Setup

```python
import numpy as np

weights = np.random.rand(100_000)
weights /= np.sum(weights)
```

## Timing

and then run

```python
In [10]: %time np.random.choice(np.random.rand(100_000), p=weights)
CPU times: user 1.3 ms, sys: 873 µs, total: 2.18 ms
Wall time: 1.43 ms
Out[10]: 0.25297089985681676
```

and then I estimated a 20MB wav file, which on SSD is ~10ms, hence around an order of magnitude

mostly I think 100K weighted choice is unlikely. but it was a fun exercise to implement either way!

worldveil:
@iver56 anything else you'd like looked at or changed to merge this?

iver56:
Thanks for the patience! I've had a look at the code now, and it looks mostly good. However, I ran into a few test failures when running the tests locally. I tried with both the numpy 1.x env and the numpy 2.x env. I noticed that CircleCI doesn't run on forks, so I guess I should improve that in the future.

Meanwhile, can you please check these test failures?

__________________________________________________________ test_one_of_weights __________________________________________________________

    def test_one_of_weights():
        samples = np.array([1.0, 0.5, -0.25, -0.125, 0.0], dtype=np.float32)
        sample_rate = 16000

        transforms = [
            Gain(min_gain_db=6.0, max_gain_db=6.0, p=1.0),  # Gain approx 2.0
            Gain(min_gain_db=-6.0, max_gain_db=-6.0, p=1.0), # Gain approx 0.5
        ]
        weights = [0.8, 0.2]

        augmenter = OneOf(transforms=transforms, p=1.0, weights=weights)

        counts = [0, 0]
        num_runs = 2000
        for _ in range(num_runs):
            perturbed_samples = augmenter(samples=samples, sample_rate=sample_rate)
            # Check which transform was applied based on the gain effect
            if np.allclose(perturbed_samples[0], samples[0] * 2.0, atol=1e-5):
                counts[0] += 1
            elif np.allclose(perturbed_samples[0], samples[0] * 0.5, atol=1e-5):
                counts[1] += 1
            else:
                 # This should not happen if p=1.0 and gains are fixed
>                raise AssertionError("Unexpected output sample value")
E                AssertionError: Unexpected output sample value

tests\test_one_of.py:222: AssertionError
___________________________________________________ test_one_of_weights_normalization ___________________________________________________

    def test_one_of_weights_normalization():
        samples = np.array([1.0], dtype=np.float32)
        sample_rate = 16000

        transforms = [
            Gain(min_gain_db=6.0, max_gain_db=6.0, p=1.0),
            Gain(min_gain_db=-6.0, max_gain_db=-6.0, p=1.0),
        ]
        # Weights don't sum to 1, should be normalized to [0.8, 0.2]
        weights = [4, 1]
        expected_normalized_weights = [0.8, 0.2]

>       augmenter = OneOf(transforms=transforms, p=1.0, weights=weights)

tests\test_one_of.py:240:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = OneOf([
  Gain(p=1.0, min_gain_db=6.0, max_gain_db=6.0),
  Gain(p=1.0, min_gain_db=-6.0, max_gain_db=-6.0),
], p=1.0)
transforms = [Gain(p=1.0, min_gain_db=6.0, max_gain_db=6.0), Gain(p=1.0, min_gain_db=-6.0, max_gain_db=-6.0)], p = 1.0, weights = [4, 1]

    def __init__(self, transforms, p: float = 1.0, weights: Optional[list[float]] = None):
        super().__init__(transforms, p)
        self.transform_index = 0
        self.should_apply = True

        # Handle weights initialization and validation
        if weights is not None:
            if len(weights) != len(transforms):
                raise ValueError("Length of weights must match length of transforms")
            if any(w < 0 for w in weights):
                raise ValueError("All weights must be non-negative")
            if sum(weights) == 0:
                raise ValueError("Sum of weights must be greater than 0")
            # Normalize weights to sum to 1
            self.weights = np.array(weights)
>           self.weights /= np.sum(self.weights)
E           numpy._core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'divide' output from dtype('float64') to dtype('int64') with casting rule 'same_kind'

audiomentations\core\composition.py:293: UFuncTypeError
_________________________________ TestWeightedChoiceSampler.test_init_valid_weights_needs_normalization _________________________________

self = <tests.test_sampling.TestWeightedChoiceSampler object at 0x000002A2EC776860>

    def test_init_valid_weights_needs_normalization(self):
        weights = [1, 4] # Sums to 5
>       sampler = WeightedChoiceSampler(weights=weights)

tests\test_sampling.py:18:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <audiomentations.core.sampling.WeightedChoiceSampler object at 0x000002A281852020>, weights = array([1, 4])

    def __init__(self, weights: list[float] = None):
        assert weights is not None and len(weights) > 0, "Weights must be provided"

        # ensure weights are normalized
        weights = np.array(weights)
>       weights /= np.sum(weights)
E       numpy._core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'divide' output from dtype('float64') to dtype('int64') with casting rule 'same_kind'

audiomentations\core\sampling.py:17: UFuncTypeError

worldveil:
Hm, they were all passing when I ran locally. You're just running the command `pytest` right?

iver56:
Yes, I ran just `pytest` (on Windows)

Btw, I just made the switch from CircleCI to Github Actions:

https://github.com/iver56/audiomentations/pull/390

I think that should make it easier to run the CI in your fork/PR. If you're interested in trying that, you can rebase your branch

PR #393, opened 2025-04-25
Fix test failures related to OneOf weighting

iver56:


worldveil:
@iver56 thanks for tackling these!

for these types of assertions
https://github.com/iver56/audiomentations/commit/7567d7dc3e9f4383bf94c7ecbdc746a7307087f8#diff-8520f0d748c797ad52f7f72cf05073177776fe1c66b5ac82c4ff64a9936e2e1aL216-L220

I usually mess with `rtol=` instead of changing the values to `0.5123498` or whatever. but glad it's working now!

When do you usually bump the pip version? Jus curious.

iver56:
Usually every 2-3 months. If you want to install the bleeding edge version already today, you can use the github link instead of the PyPI package when installing audiomentations.

PR #398, opened 2025-06-09
Added overdrive effect

worldveil:
I found this effect in an old PR to this repo, and liked the sound of it!
https://github.com/iver56/audiomentations/pull/244

Added now with tests.


iver56:
Thanks for sharing! I see two things that make it a bit harder to get this in:

* This PR adds two transforms. It would be easier to review them individually.
* One of them depends on torch & torchaudio. They are large, but in reality many already have it installed anyway, so it might not be a big problem. But the dependencies must be optional, and imported on demand.

PR #418, opened 2026-01-14
Update repeat_part.py

lubacien:
Fixing bug happening if we go too far to the right of the audio


iver56:
Thanks for the PR! Could you add a test that reproduces the original issue that you are patching here?

iver56:
@lubacien Reminder for the request above ^

PR #419, opened 2026-01-17
Add shared audio cache to accelerate large, parallel jobs

worldveil:
## Summary

Adds a pluggable cache backend interface to `load_sound_file()` that enables sharing loaded/resampled audio across parallel workers. This eliminates redundant disk I/O and resampling when processing large datasets with frameworks like Ray.

I provide a small guide that shows how to make an LRU cache with configurable memory size using Ray, but note that no dependency for Ray is needed for `audiomentations`, and that other object stores are usable as well.

## Motivation

Transforms like `AddBackgroundNoise` and `ApplyImpulseResponse` load audio files from disk and resample them on each use. In multiprocessing each worker would load and resample these files individually, slowing down the job since each worker may be duplicating effort and memory.

## Solution

A simple cache backend interface with `get(key)` and `put(key, audio)` methods:

```python
from audiomentations import set_audio_cache_backend

class MyCacheBackend:
    def get(self, key): ...  # Returns np.ndarray or None
    def put(self, key, audio): ...

set_audio_cache_backend(MyCacheBackend())
```

Key format: `(file_path, sample_rate, mono)`

## Key behaviors

Cache miss: Loads full file, caches it, then slices if needed for offset/duration
Cache hit: Returns copy from cache, slices if needed

No audiomentations dependencies: Backend implementation is user-provided (can be Ray Actor, Redis, dict, etc.)

## Changes

- `audiomentations/core/audio_loading_utils.py`: Added set_audio_cache_backend(), get_audio_cache_backend(), clear_audio_cache_backend()
- `audiomentations/__init__.py`: Export new functions
- `tests/test_shared_audio_cache.py`: Tests covering cache hit/miss, slicing, transforms
- `docs/guides/shared_audio_cache.md`: Usage guide with Ray Actor example

### DIFF
--- a/audiomentations/__init__.py
+++ b/audiomentations/__init__.py
@@ -38,5 +38,10 @@ from .augmentations.time_mask import TimeMask
 from .augmentations.time_stretch import TimeStretch
 from .augmentations.trim import Trim
 from .core.composition import Compose, OneOf, SomeOf
+from .core.audio_loading_utils import (
+    set_audio_cache_backend,
+    get_audio_cache_backend,
+    clear_audio_cache_backend,
+)
 
 __version__ = "0.43.1"
--- a/audiomentations/core/audio_loading_utils.py
+++ b/audiomentations/core/audio_loading_utils.py
@@ -3,20 +3,57 @@ import librosa
 import numpy as np
 
 
-def load_sound_file(file_path, sample_rate, mono=True, resample_type="auto", offset = 0.0, duration = None):
-    """
-    Load an audio file as a floating point time series. Audio will be automatically
-    resampled to the given sample rate.
+# =============================================================================
+# Audio Cache Backend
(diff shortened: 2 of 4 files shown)


iver56:
Thanks for sharing! This looks useful when you have many workers, lots of memory and a moderately sized set of sounds. If a significant amount of people want this feature, I could consider merging this. Note: Unfortunately, I do not have much bandwidth to work on open source projects these days.

worldveil:
Yep, precisely those conditions! If sounds set for augmentations is too large there's not much point. 

The example guide I wrote shows how to restrict the size of the shared LRU cache by a certain size of memory, so one can tune it to your cache access patterns & memory budget. 

Alright, I'll link to this PR from some issues and if there's momentum, very happy to help merge it.

